PR #39, opened 2026-06-11
Harden render hot path, session layer, and URL output (perf/memory/security)

pal-tamas:
Top-impact sweep across performance, memory, and security, scoped to fixes that run on every render/diff or close a real DoS/XSS gap. Full non-e2e suite green; `Rask.Example.Wasm` still trims warning-free.

## Security
- **URL-scheme sanitization by default** (was a live DOM-XSS hole). HTML-encoding alone leaves `<a href="javascript:…">` executable. New `UrlSanitizer` + `Component.AppendUrlAttr`/`AppendMediaUrlAttr` neutralize `javascript:`/`vbscript:`/untrusted `data:` to `about:blank` across every URL-bearing tag (`href`, `src`, `cite`, `formaction`, object `data`, `poster`, SVG `href`). Detection defeats leading-whitespace / embedded tab/newline/NUL obfuscation. Per-call opt-out via `RaskUrl.Trusted(...)` (still HTML-encoded). Media tags still allow inline `data:image|video|audio`.

## Performance
- **Lock-free scoped-asset reads** — the by-`Type` lookups hit per component per render move to `ConcurrentDictionary`, dropping the process-wide lock that serialized concurrent sessions.
- **`AsyncLocal` off the hot path** — `[ThreadStatic]` `CurrentSync` mirror for the synchronous render walk; per-element attribute reads use it (lifecycle reads stay on `AsyncLocal`). Cached `Component.Key` stringification (no per-render `ToString` alloc on keyed lists).
- **Head splice** — `ApplyTo` takes the pre-computed sentinel index (no second whole-body `IndexOf`), pools its `StringBuilder`s, appends keys instead of per-asset `string.Insert`.

## Memory
- **Hoisted render-context collections** — `HeadAssetRegistry` + `MountedTypes` live on the root `LiveState`, reused and cleared per render instead of allocating ~5 collections every frame.
- **Bounded session retention + hard cap (DoS)** — a GET-minted session that never sends `hello` is evicted on a short 10s grace (vs the 30s reconnect grace); `MaxSessions` is now an atomic `TryCreate` reservation taken before the tree is built, so a concurrent GET burst can't race past the cap.
- **Handler-chain backpressure (DoS)** — a per-session pending-dispatch counter (`MaxPendingHandlers`, default 512) closes the socket when the chain backs up behind a hung/flooding client, instead of retaining cloned `JsonElement`s without bound.

## Tests
- `UrlSanitizerTests` (26): dangerous/obfuscated schemes neutralized, safe URLs pass through, `data:image` allowed on media, `RaskUrl.Trusted` round-trips, still HTML-encoded.
- `LiveSessionStoreTests`: `TryCreate` hard cap incl. a 100-way concurrent burst against a cap of 10.
- `HandlerBackpressureTests`: hung-handler flood closes the socket; normal traffic stays open.
- `HeadAssetRenderTests`: repeated renders produce byte-identical head (guards the reused/cleared registry).

## Notes / deviations from plan
- A1 uses `ConcurrentDictionary` rather than a `FrozenDictionary` COW — same lock-free-read goal without an O(n²) startup rebuild.
- A3 implements the safe wins (sentinel-index reuse, pooled builders, append-based keying, B1 collection hoist) but **omits** the speculative content-hash memoization of the whole head block: head asset strings are re-rendered each frame regardless, so the memo's marginal benefit didn't justify the risk to the delicate head-morph reconciliation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,32 @@ them until tagged releases begin.
 
 ## [Unreleased]
 
+### Security
+- URL-bearing attributes (`href`, `src`, `cite`, `formaction`, object `data`, `poster`,
+  SVG `href`) are now **scheme-sanitized by default**: `javascript:`/`vbscript:` and
+  `data:` outside media tags are neutralized to `about:blank`, closing a DOM-XSS hole that
+  HTML-encoding alone left open. Detection defeats whitespace/tab/NUL obfuscation. Opt out
+  per call with `RaskUrl.Trusted(...)` for URLs you control; media tags still allow inline
+  `data:image/*`, `data:video/*`, `data:audio/*`. See the [getting-started guide](docs/getting-started.md#url-attributes-are-scheme-sanitized).
+
+### Performance
+- Scoped-asset registry reads (run per component, per render) are now lock-free
+  (`ConcurrentDictionary`), so concurrent sessions no longer serialize on a process-wide lock.
+- Removed `AsyncLocal` reads from the per-element attribute path via a thread-local
+  render-context mirror, and cache `Component.Key` stringification (no per-render `ToString`
+  allocation on keyed lists).
+- `<head>` splice avoids a second whole-body scan, pools its builders, and appends keys
+  without per-asset string allocation.
+
+### Memory
+- The per-render head-asset collector and mounted-type set are hoisted onto the root and
+  reused (cleared per render) instead of allocating fresh collections every frame.
+- A session minted by the GET shell but never connected over WebSocket is now evicted on a
+  short grace (vs. the 30s reconnect grace), and `MaxSessions` is enforced as a hard atomic
+  reservation — a concurrent GET burst can no longer exceed the cap.
+- The WebSocket handler-dispatch chain is bounded: when handlers back up behind a hung or
+  flooding client, the socket is closed instead of retaining queued payloads without limit.
+
 ---
 
 ## [0.7.0] - 2026-06-10
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -91,6 +91,24 @@ P()["<b>encoded</b> — shows the angle brackets as text"],   // string → Text
 Div()[Raw("<b>bold</b>")]                                    // emitted verbatim
 ```
 
+### URL attributes are scheme-sanitized
+
+URL-bearing attributes — `href` (`A`, `Area`, `Link`, `Base`, SVG), `src`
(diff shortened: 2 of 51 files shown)


pal-tamas:
## Benchmark results

Apple M4 Pro (Arm64), .NET 10.0.5, BenchmarkDotNet 0.14.0, `--job short`. `main` (baseline) vs this branch, same filters. Full-render benchmarks use `InvocationCount=1` (their `[IterationSetup]` rebuilds the tree) so their `Mean` carries large error — the deterministic **Allocated** column is the reliable signal there.

### Headline — scoped-asset registry read (A1: lock-free reads)

`AssetLoadingBenchmarks.TryGetScopeId_200Lookups` — the lookup run per component per render:

| | Mean | Allocated |
|---|---:|---:|
| main | 1.693 µs | 0 B |
| branch | **0.543 µs** | 0 B |

**~3.1× faster** (67% less time), zero allocation — the process-wide lock that serialized concurrent sessions is gone.

### Render path with scoped CSS (A1 + A2)

`HtmlSerializerBenchmarks.RenderTreeWithScopedCss` (200-row scoped-CSS tree):

| | Mean | Allocated |
|---|---:|---:|
| main | 23.48 µs | 33.09 KB |
| branch | **19.22 µs** | **32.58 KB** |

**~18% faster**, slightly less allocated. `EmitMountedAssets_50` likewise ~14% faster (1.65 → 1.42 µs; output bytes unchanged).

### Full live-render round trip (A2 / A3 / B1) — alloc, no regression

`LiveRenderRoundTripBenchmarks`, **Allocated** (Mean omitted — error > mean at `InvocationCount=1`):

| Scenario | main | branch |
|---|---:|---:|
| RenderOnce | 89.45 KB | 93.23 KB |
| RenderTenTimes | 162.72 KB | 163.54 KB |
| RenderKeyedList100_ShuffledEachIteration | 114.05 KB | 119.49 KB |
| RenderDeep_50UserComponents | 79.74 KB | 83.41 KB |

These are flat within run-to-run variance. An earlier cut of the URL sanitizer (`new string(...)` per `href`/`src` per render) showed a real ~8 KB regression here (`RenderTenTimes` 171 KB); the follow-up commit (`660db92`) made the scheme check span-based and allocation-free, bringing it back in line — included in the numbers above.

### Not measured
`payload-bytes` (wire size) is unchanged — none of these changes touch the diff codec output, so the established baselines in `benchmarks/Rask.Benchmarks/Baselines` still hold.

PR #61, opened 2026-06-12
refactor!: remove Component.User; + EF Core sample + session-removal race fix

pal-tamas:
## Summary
Branch bundles three changes destined for `main`:
- **`refactor!: remove Component.User`** (`9d2d781`) — drop the `Component.User` member; inject `IUserProvider` instead. **Breaking.**
- **`feat(samples): EF Core + SQLite CRUD sample`** (`0c616ae`).
- **`fix(server): avoid ObjectDisposedException retiring pending session removals`** (`0e11c04`) — `ScheduleRemoval` disposed the prior pending-removal `CancellationTokenSource` from inside an `AddOrUpdate` factory (while still reachable through `_pendingRemovals`), so a concurrent `CancelPendingRemoval`/`CancelAllPending` (two sockets detaching at shutdown) could `TryRemove` the same instance and `Cancel()` an already-disposed source. Now swaps atomically via `TryUpdate`/`TryAdd` and retires the prior source only after the CAS makes it unreachable, so exactly one thread owns its disposal.

## Testing
- Unit: `Rask.Server.Tests` — **163/163 passed**. `Reconnect_WhileExistingSocketAttached_NewSocketBecomesAuthoritative` passed 5/5 consecutive runs (was flaky on the dispose race).
- `dotnet build -warnaserror` (Rask.Server) — clean; `dotnet format` on `LiveSessionStore.cs` — clean.
- E2E: `samples/` change is the EF Core sample (`feat` commit, pre-existing on branch); E2E matrix runs in CI.

## Benchmarks
n/a — the fix is in the session-removal scheduler, not the render/runtime hot path.

## CHANGELOG
- [Unreleased] → Fixed: `LiveSessionStore` no longer throws `ObjectDisposedException` while retiring pending session removals under contention.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,6 +83,11 @@ ehthumbs.db
 Desktop.ini
 $RECYCLE.BIN/
 
+# EF Core SQLite sample database (created at runtime by samples/Rask.Example.EfCore)
+raskExampleCatalog.db
+raskExampleCatalog.db-shm
+raskExampleCatalog.db-wal
+
 # Coverage
 *.coverage
 *.coveragexml
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,31 @@ them until tagged releases begin.
 
 ## [Unreleased]
 
+### Changed
+- **WASM build migrated to `Microsoft.NET.Sdk.WebAssembly`** so framework assets are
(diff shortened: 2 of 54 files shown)


pal-tamas:
Superseded: the Component.User removal already landed on main via #58; the EF Core sample and session-removal fix landed via #59 and #62. Closing to avoid re-applying the already-merged breaking change. Branch retained (holds local WIP).

PR #80, opened 2026-06-16
feat(assets): eager-preload scoped CSS/JS to eliminate navigation FOUC

pal-tamas:
## Summary

The page `<head>` now also emits a non-blocking `<link rel="preload" fetchpriority="low">` for **every** registered scoped asset — not just the components on the current route. When a component first mounts later (client-side navigation, a conditionally rendered section), its scoped stylesheet/script is already in the browser's HTTP cache, so:

- the body swaps with **no flash of unstyled content** (the client FOUC guard's `performance.getEntriesByName` check passes → no 500ms stall), and
- the scoped-JS `window.Rask[Type]` namespace is ready on first interaction (no cold namespace-poll latency).

Scoped CSS is selector-rewritten to `[data-r-xxxx]`, so preloading an unmounted component's styles has no visual effect until its elements exist — preload-all is visually safe.

On by default; opt out with `AddRask(o => o.PreloadScopedAssets = false)` (or the WASM host-builder equivalent).

## How it works

- New `EmitScopedPreloads` pass runs in `HeadAssetRegistry.ApplyTo` after the existing mounted, render-blocking assets (so those keep discovery priority).
- The preload block is **render-independent and cached**, rebuilt only when the asset set changes — tracked by a new monotonic `ScopedAssetRegistry.Version` (bumped on every register/unregister/invalidate; `AssetChanged` intentionally does not fire on the bulk-clear paths, so a whole-registry projection can't rely on it).
- **No client change needed**: the preload links are `rel="preload"` — they match neither the JS invoke gate (`tagName === "SCRIPT"`) nor the FOUC guard (`link[rel="stylesheet"]`) in `rask.js` / `rask.wasm.js`, so they stay inert. Verified in both runtimes.

## Testing

- 10 new unit tests in `ScopedPreloadEmissionTests` (emission shape, non-blocking attrs, CSS-before-JS order, PathBase prefixing, cache reuse, cache rebuild on registry mutation, null-arg, and a real-pipeline `ApplyTo` integration test through `LiveRenderContext` + `HtmlSerializer`).
- Full unit suite green: Core 1492, Server 170, Wasm.Tasks 10, Example.Shared 252.
- Existing `EmitMountedAssets` tests unchanged (the pure mounted path was left untouched).

## Benchmarks

New `AssetLoadingBenchmarks` cases (`MemoryDiagnoser`):

| Method | Mean | Allocated |
|---|---:|---:|
| `EmitMountedAssets_200` (baseline) | 1,658 ns | 19.12 KB |
| `EmitScopedPreloads_Warm` | 834 ns | 16.07 KB |
| `EmitMountedAssetsWithPreloads_200` | 2,477 ns | 34.81 KB |

`EmitScopedPreloads_Warm`'s 16.07 KB is essentially the benchmark's own `new StringBuilder(8192)` (8192 chars × 2 = 16 KB); the cached-block lookup (an `Interlocked.Read` + reference compare) and single append add no measurable heap. In production `ApplyTo` appends into the already-pooled head builder, so the real per-render cost is just the emitted markup bytes — no registry snapshot, no rebuild.

## Changelog

Added under `[Unreleased]`.

## Follow-ups (tracked, not in this PR)

- Switch off-route hints to `rel="prefetch"` (avoids Chrome's "preloaded but not used" warning + lower priority).
- `103 Early Hints` for the current route's critical CSS (Server mode).
- Inline critical scoped CSS for the initial route; pre-compress (Brotli) served assets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ them until tagged releases begin.
 
 ## [Unreleased]
 
+### Added
+- **Eager preload of scoped CSS/JS eliminates navigation FOUC.** The page `<head>` now also
+  emits a non-blocking `<link rel="preload" fetchpriority="low">` for *every* registered scoped
+  asset — not just the components on the current route — so when a component first mounts later
+  (client-side navigation, a conditionally rendered section) its stylesheet/script is already in
+  the browser cache: the body swaps with no flash of unstyled content and the scoped-JS namespace
+  is ready on first interaction. The preload markup is render-independent and cached (rebuilt only
+  when the asset set changes), so it costs a single append per render. On by default; opt out with
+  `AddRask(o => o.PreloadScopedAssets = false)` (or the equivalent WASM host-builder option) to
+  fetch each scoped asset only when its component first mounts.
+
 ### Fixed
 - **Switching a syntax-highlighted code-sample tab no longer renders the new pane's markup as
   literal text.** A live re-render that replaced one `Raw` value with another (e.g. switching a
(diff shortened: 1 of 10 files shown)


pal-tamas:
Superseded by #82 (squashed #80 + #81 into a single PR).

PR #81, opened 2026-06-16
refactor(assets): use rel=prefetch for off-route scoped asset warming

pal-tamas:
**Stacked on #80** — base is `feat/eager-preload-scoped-assets`, so this diff shows only the prefetch change. Retarget to `main` once #80 merges.

## Why

#80 emits `<link rel="preload" fetchpriority="low">` for *every* registered scoped asset. Most are off-route and may never be used this session, so:

- Chrome logs **"resource was preloaded but not used within a few seconds"** for each off-route asset, and
- `preload` competes at a higher base priority than these future-navigation hints deserve.

## What

Switch the cached block to **`<link rel="prefetch">`** (dropping the now-redundant `fetchpriority="low"` — prefetch is already the lowest-priority hint). `prefetch` is the textbook future-navigation hint:

- no "unused preload" console warning,
- never competes with the current route's critical resources,
- still lands the bytes in the HTTP cache so a later mount is flash-free.

`as="style"/"script"` is kept so the browser sets the right Accept header and cache partition. The current route's mounted assets still ship as render-blocking `<link>`/`<script>` via `EmitMountedAssets`; a coincident low-priority prefetch for the same href coalesces.

## No client change

The links remain inert to both client runtimes: `rel="prefetch"` matches neither the JS invoke gate (`tagName === "SCRIPT"`) nor the FOUC guard (`link[rel="stylesheet"]`). The FOUC guard's warm-cache check (`performance.getEntriesByName`) still sees the prefetched resource-timing entry, so the no-FOUC behavior holds.

## Tradeoff (noted for the record)

`preload` is a slightly *stronger* no-FOUC guarantee (kept in the memory cache, immediately usable); `prefetch` is best-effort and may be evicted under memory pressure or stored in a shorter-lived cache. For Rask's in-place morph navigation the future-navigation semantics + no console noise of `prefetch` are the better default, but the eviction nuance is real — happy to revisit if you'd rather keep `preload`.

## API

The public `PreloadScopedAssets` option keeps its name (it's the feature toggle); only the emitted hint and its `data-rask-key` (`rsk-prefetch-*`) change.

## Testing

- `ScopedPreloadEmissionTests` updated to assert `rel="prefetch"` / `rsk-prefetch-*` and renamed accordingly; all 10 pass.
- Full Core suite green (1492).
- Per-render cost is unchanged from #80 — still a single append of the cached block (the benchmark cases call the same `EmitScopedPreloads`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,15 +8,17 @@ them until tagged releases begin.
 ## [Unreleased]
 
 ### Added
-- **Eager preload of scoped CSS/JS eliminates navigation FOUC.** The page `<head>` now also
-  emits a non-blocking `<link rel="preload" fetchpriority="low">` for *every* registered scoped
-  asset — not just the components on the current route — so when a component first mounts later
-  (client-side navigation, a conditionally rendered section) its stylesheet/script is already in
-  the browser cache: the body swaps with no flash of unstyled content and the scoped-JS namespace
-  is ready on first interaction. The preload markup is render-independent and cached (rebuilt only
-  when the asset set changes), so it costs a single append per render. On by default; opt out with
-  `AddRask(o => o.PreloadScopedAssets = false)` (or the equivalent WASM host-builder option) to
-  fetch each scoped asset only when its component first mounts.
+- **Eager prefetch of scoped CSS/JS eliminates navigation FOUC.** The page `<head>` now also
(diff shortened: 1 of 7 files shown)


pal-tamas:
Superseded by #82 (squashed #80 + #81 into a single PR).

PR #124, opened 2026-06-22
feat(server): cap inbound WebSocket frame rate per connection

pal-tamas:
## Summary

Adds a per-connection **inbound WebSocket frame-rate cap** to the Server receive loop, closing a
CPU-DoS gap the existing bounds miss.

`MaxInboundFrameBytes` (8 MB) caps a single frame's **size**, and `MaxPendingHandlers` (512) bounds
queued **handler** dispatches — but neither bounds the **rate** of small *non-handler* frames
(`jsResult` / `navigate` / `dotNetInvoke` / malformed). Each of those still costs a `JsonDocument`
parse, so a high-rate stream of them could burn CPU unbounded on a single connection.

The receive loop now counts every completed receive over a sliding one-second window; past
`MaxInboundFramesPerSecond` it closes the socket with a policy-violation status (same graceful-close
shape as the handler-backlog breaker), and the client reconnects against the intact session and
resumes from current state. The count happens **before** the per-frame JSON parse, so the parse cost
itself is bounded.

### Design
- Fixed internal default of **1000 frames/s** — far above any legitimate interaction burst (rapid
  typing with per-keystroke handlers, or a 60 Hz scroll handler, is well under 100/s). Like the
  sibling caps (`MaxInboundFrameBytes`, `MaxPendingHandlers`) it's an internal default, **not a public
  knob**: operators tune ingress with a reverse-proxy / WAF rate limit, which this complements (it
  guards a single misbehaving connection, not connection-count or cross-connection floods).

### Docs
- `docs/authentication.md` **Hardening reference** gains a row covering all three inbound-frame
  bounds (size, rate, handler backlog) + the reverse-proxy guidance; the **Security checklist** notes
  the protections are automatic.

## Testing
- New `WebSocketFrameRateTests`: a flood (cap lowered to 5) closes the socket; under-cap traffic
  (hello + 10 frames) is processed and the socket stays open. **Both pass.**
- `WebSocketFrameSizeTests` (sibling cap) still green.
- `dotnet build src/Rask.Server -warnaserror` clean; `dotnet format --verify-no-changes` clean.
- Full `Rask.Server.Tests`: **205/206**. The one failure —
  `CheckboxChange_SetsModelToReportedState_AcrossManyClicks` — is a pre-existing coalescing-timing
  flake (it sends ~6 awaited frames, nowhere near the cap; passes in isolation; the rate check is a
  no-op increment below the cap, so this change cannot affect it).

## Benchmarks
None — the added work per frame is one `Environment.TickCount64` read + an `int` increment, off the
render/diff hot path.

## Changelog
`[Unreleased] → Security` entry added.


pal-tamas:
## Correction + flake fix (commit 75091bb)

My original description called the `CheckboxChange_…AcrossManyClicks` failure a *pre-existing*
flake. That was wrong — **this PR's new test caused it**, and it's now fixed.

**Root cause.** `WebSocketFrameRateTests` sets the *static* `MaxInboundFramesPerSecond = 5` to
exercise the breaker, but it was in no xUnit collection, so it ran in parallel with other tests and
the process-global cap leaked onto their connections. `CheckboxBindingDiffTests` sends 6 frames
(hello + 5 clicks); with the cap held at 5 by the concurrent rate test, its 6th frame tripped the
breaker and the server closed the socket — failing at "click 4".

Deterministic proof (cap held at 5, capturing every frame):

```
frame 1 (hello):        state=Open          S=False
frame 2 (click1 true):  state=Open          S=True
frame 3 (click1 true'): state=Open          S=True
frame 4 (click2 false): state=Open          S=False
frame 5 (click3 true):  state=Open          S=True
frame 6 (click4 false): state=CloseReceived <null/closed>   ← 6th frame trips cap=5
```

**Fix.** Put both static-cap Security tests (the rate test + the sibling `WebSocketFrameSizeTests`,
which mutates `MaxInboundFrameBytes`) in `[Collection("SessionGracePeriod")]`
(`DisableParallelization`) — the same pattern `HandlerBackpressureTests` (which sets
`MaxPendingHandlers`) and the `DiffMode` flippers already use — so the global mutation can't overlap
another test's traffic.

**Verification (A/B over the two implicated classes):**
- Without the guard: **20/20** runs failed.
- With the guard: **0/20** failed.
- Full `Rask.Server.Tests`: **206/206**.

PR #334, opened 2026-07-13
fix(core): preserve foreign head nodes across re-renders

pal-tamas:
## Summary
Makes Rask's `<head>` reconciliation **preserve foreign (third-party-injected) head nodes** instead of trimming them on re-render. Libraries routinely inject `<style>`/`<link>`/`<script>` into `<head>` at runtime (Monaco's theme colours, Chart.js, syntax highlighters, analytics); previously the live-diff morph removed any such node on the next re-render (it isn't in the .NET-rendered head), so e.g. a mounted code editor lost all its styling on the first state change.

The client morph (`rask-morph.js`, shared by Server/WASM/Native) now tags the head nodes it owns (`__raskHead`) and preserves any head element it doesn't (nodes that appeared since the last render) — the same effect as `data-rask-managed`, but automatic. A `raskHeadReconciled` gate keeps the **first** head reconciliation (boot-shell hydration: importmap / `<base>` / preload / scoped placeholders) byte-identical to before; preservation applies only to nodes injected *after* a render. Scoped to `<head>` — body reconciliation is unchanged.

This is the framework-level version of the playground's Monaco head-strip fix (#333), which is dropped in favour of the framework doing it for everyone.

## Testing
- E2E: **PlaygroundExampleTests** — pass, with the app-level Monaco head-guard removed, so it now validates the framework fix (asserts Monaco's head `<style>` survives Run). Verified before/after Run in both **Chromium and WebKit** (Safari's engine): token stays `rgb(86,156,214)`, head style count stays 1.
- E2E regression: **StandaloneWasmExampleTests** (the comprehensive WASM journey — every page, scoped-CSS delivery, shell reload, live-ticker publish-render) — pass, confirming no head-hydration regression.
- `dotnet format` clean; full solution builds warnings-as-errors clean; materialized WASM/Native client bundles regenerated.
- The full CI E2E matrix (Server, WASM, SubPath, Standalone, Playground, Auth×4) exercises head hydration across every host.

## Benchmarks
n/a — the change is client-side DOM morph JS (not the .NET render/diff hot path the benchmarks measure). Added work is an O(head-children) marking pass (~7–10 nodes) per head morph.

## CHANGELOG
- [Unreleased] → Changed: "Head reconciliation preserves foreign (third-party-injected) head nodes."

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -189,6 +189,16 @@ them until tagged releases begin.
   over the WebSocket.
 
 ### Changed
+- **Head reconciliation preserves foreign (third-party-injected) head nodes.** Libraries routinely inject
+  `<style>`/`<link>`/`<script>` into `<head>` at runtime — Monaco's theme colours, Chart.js, syntax
+  highlighters, analytics. Previously the live-diff morph trimmed any such node on the next re-render (it
+  wasn't in the .NET-rendered head), so e.g. a mounted code editor lost all its styling on the first state
+  change. The client morph now tags the head nodes it owns (inline, as each rendered node is placed) and
+  leaves any it doesn't (nodes injected since the last render) in place — the same effect `data-rask-managed`
+  gives, but automatic, so a third-party library that injects into `<head>` "just works". The first head
+  reconciliation (boot-shell hydration) is unchanged, and `data-rask-key` nodes (the framework's own keyed
+  head links, incl. the scoped-CSS FOUC preload clone) always reconcile by key rather than duplicate. This
+  let the playground drop its bespoke Monaco head-guard.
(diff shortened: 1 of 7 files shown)


pal-tamas:
Closing in favour of the existing app-level mechanism. Two adversarial review rounds showed the morph-level head-preservation change destabilises load-bearing code — most seriously it broke conditional head rendering (diff-inserted <head> nodes weren't marked owned) and added a per-node allocation on the render hotpath. The framework already provides the correct, targeted opt-out (`data-rask-managed`, which the playground's #333 fix uses), so the playground works today with zero risk to the reconciler. Follow-up: document the third-party-head-injection pattern.

PR #385, opened 2026-07-15
feat(sqlite): load harness for Rask.SQLite + EF Core, and fix a lost retry clock

pal-tamas:
## Summary

`docs/sqlite.md` made strong, specific claims — WAL means readers don't block the writer, a constant 1 ms fair interval beats backoff on tail latency, `journal_size_limit` caps the WAL at 64 MiB — and **none of them were measured**. What existed proved correctness, not performance: `SqliteConcurrencyStressTests` (200/500 writers, no timing) and `SqliteWriteContentionBenchmarks` (BenchmarkDotNet, raw path only, a burst's mean). The EF Core path had **no multi-writer coverage at all**.

This adds `benchmarks/Rask.Benchmarks.Sqlite`, a load harness that reports sustained throughput, tail latency and error rates — which BDN structurally cannot. It is a plain exe, split out of `Rask.Benchmarks` (as `Rask.Benchmarks.VsBlazor` is) because it needs EF Core and because a minutes-long soak has no place in a filtered BDN sweep. The existing BDN suite stays; it answers per-operation cost.

## It found a real bug (`### Fixed`)

`RaskSqliteExecutionStrategy` started its retry clock on the first contention and **never released it**, and EF Core hands one `DbContext` a single strategy instance for its whole lifetime. Once `Timeout` of wall-clock had passed since that first contention, every later `SaveChanges` on that context gave up with `database is locked` **without a single retry** — the failure got *more* likely the longer a context lived, and a scoped context under load outlives 5 s easily. Fixed by resetting the clock in `OnFirstExecution`, with a regression test that fails without the fix.

## What the numbers say

Full tables in [`docs/sqlite.md`](docs/sqlite.md#load-test-numbers); every figure reconciles against the committed `Baselines/sqlite-load.csv`.

- **~99k mixed 90/10 ops/s at p99 10 ms** on one file (~26k through EF Core) — the OPF "one server is enough" number.
- **WAL is worth ~95–154×** read throughput under a concurrent writer. Measured against a `DELETE`-mode control arm, because "WAL readers are fast" is meaningless without the journal mode it replaced.
- **The fair-interval retry does not buy throughput — it buys a bounded worst case.** Uncontended it is ~21% *slower* than plain `busy_timeout`, whose p50 and p99 also stay better under load because its distribution is bimodal. What it changes is the tail: **174 ms vs ~16 s max** (~92×). Documented honestly rather than spun.
- **`journal_size_limit` cannot cap WAL growth while a leaked read transaction pins it**: 3.16 GB in 90 s against a 0.5 MiB database. It truncates *after* a checkpoint, so it can't act while one can't run. Correct SQLite behaviour, and a disk-space incident waiting to happen.

## The gate

`check` asserts **invariants and same-run ratios, never absolute milliseconds** — `ci.yml` already rules timing out as "too noisy on shared runners to gate on", and a flaky gate gets disabled within a week. Every comparison is between two arms in the same process on the same box, so a 2-vCPU runner and an M4 agree on it. Tier 1 (any hardware): no escaped `SQLITE_BUSY`, no lost writes, readers unblocked, the `DELETE` control still slow, and the non-blocking path still bounding the tail. Tier 2 (real box): EF never blocking for the driver's default command timeout.

`Baselines/README.md` states plainly what it does **not** catch: a 20% throughput regression. Nothing cheap does.

## Testing

- Full local gates green: `scripts/run-unit-local.sh` (format + warnings-as-errors + 28 test projects) and `scripts/run-e2e-local.sh` (25/25).
- `scripts/run-sqlite-load-local.sh` — 16/16 invariants pass.
- **The gate was verified to bite**: crippling the raw retry budget produced 669,328 escaped `SQLITE_BUSY`; pointing the `DELETE` control arm at WAL failed the control invariant. Both reverted.
- Graceful `Ctrl-C` verified end-to-end: results kept, CSV written, zero leaked database files, no unhandled exception.
- No new unit tests for the harness itself, matching `PayloadBytesReport`/`BundleSizeReport` precedent (benchmark tooling is verified by running it); the maths and error classifier are small static classes.

## Review fixes applied

A high-effort review caught, and this PR fixes: a published p50 column that **contradicted the baseline CSV it cited as evidence** (built from a stale console run — all tables are now machine-checked against the CSV); a forced GC sitting *between* the warmup snapshot and the clock, inflating throughput and polluting the tail the gate reads; a 4,096-entry `Random` pre-fill that turned any higher `--vus` into an error storm misreported as a SQLite failure; and a `Ctrl-C` path that discarded every measured arm and let teardown race live VUs.

## Known, documented, not hidden

An earlier `mixed-ef` @ 128 VU run escaped `SQLITE_BUSY` twice in ~542k ops, well inside its 30 s budget. The committed run is clean and it has never reproduced on the write-only arm. **Not root-caused**, recorded in `Baselines/README.md` and flagged in the docs rather than papered over.


pal-tamas:
## Root-cause attempt on the `mixed-ef` BUSY escape

**Not root-caused.** Two hypotheses falsified, no reproduction. Reporting it as open rather than closing it, but the investigation added two regression tests worth keeping on their own merit.

### The one hard constraint

The run was **15 s**; the arm's retry budget is **30 s**. The strategy cannot have exhausted its budget — it either gave up early or never saw the exception. (My earlier "max latency was 1.5 s, so it gave up early" reasoning in the baselines README was junk: BUSY ops' latencies are never recorded, so that number said nothing about the failing op. Corrected.)

Also worth weighing: that same sweep produced the *only* occurrence of the `raw-nonblocking` error burst too. Two unrelated anomalies in one run, neither seen before or since, is evidence about **that run** more than about the library.

### Ruled out (each now a committed test)

| Hypothesis | Verdict |
|---|---|
| `PRAGMA journal_mode=WAL` is lock-taking and runs on **every** open via the interceptor — and `configureRetry` sets `busy_timeout=0`, deleting the very wait that `SqlitePragmas`' busy-timeout-first ordering exists to provide (its own comment says so) | **No** — `RaskSqliteOpenUnderLockTests`. Opening and reading while another connection holds the write lock does not throw. `journal_mode=WAL` only takes the exclusive lock when it *changes* mode; on an already-WAL database it's a no-op. The ordering comment is about a brand-new database. |
| The measurement deadline cancels an op mid-retry and the harness misreads it as an escaped BUSY — "2" is suspiciously close to the ops in flight at 128 VUs | **No** — `RaskSqliteCancelledRetryTests`. A cancelled retrying `SaveChanges` throws `OperationCanceledException` at the **top level**, which the classifier scores `Cancelled`, not `Busy`. |

Both tests are keepers regardless of this bug: the second pins a real guarantee — **an aborted request mid-write must not surface as "database is locked"** — which nothing covered before.

### Not reproduced

- ~2.1M targeted `mixed-ef` @ 128 VU ops (**8× the original exposure**), 6 runs
- a full `all` sweep
- 3× (pinned-WAL soak → `mixed-ef` @ 128), recreating the original sequence

### Left for whoever sees it next

The harness now prints the exception's **extended** result code and full chain. That's the thread to pull: it separates plain `SQLITE_BUSY` from `SQLITE_BUSY_SNAPSHOT` (unretryable deferred read-then-write upgrade) and `SQLITE_BUSY_RECOVERY` (WAL/`-shm` recovery race at open) — different causes, different fixes. Without that code, any further guessing is unfalsifiable.

Gates re-run green: full unit gate, and the load gate at 16/16. (One `Rask.Outbox.Tests` failure appeared under load and passed 3/3 in isolation — `Rask.Outbox` has no reference to `Rask.SQLite` at all, so it's a pre-existing timing flake, unrelated.)

PR #434, opened 2026-07-16
feat(cli): rask deploy — one-command deploy to a single host over SSH

pal-tamas:
## Summary

Adds `rask deploy` — the roadmap's flagship missing CLI verb. It builds the app's Docker image on a single host and runs it, **entirely over SSH**: every step is `docker -H ssh://<host> …`, so there's no registry, no local Docker daemon, and no image tarball — the build context ships to the host's daemon and builds there. It deploys the `--docker` Dockerfile the templates already scaffold.

```bash
rask deploy --host deploy@box --domain app.example.com   # → https://app.example.com, auto-HTTPS
rask deploy --host deploy@box --port 8080                 # no domain: publish a port, bring your own TLS
rask deploy                                               # redeploy — host/domain remembered
rask deploy --dry-run --host deploy@box --domain app.example.com
```

**With `--domain`:** a shared **Caddy** reverse proxy fetches an automatic **Let's Encrypt** certificate — one command to a live HTTPS site. Deploys are **zero-downtime** (blue-green): the new container starts alongside the old, is health-checked, then Caddy is reloaded to point at it before the old one is retired; a container that fails to start leaves the previous version serving.

**Multiple apps share one host:** each container carries `rask.*` labels, so the proxy's routing is regenerated from the host's live containers on every deploy — a second `--domain` never disturbs the first. Without `--domain`, the app is published on `--port` (default 8080) for your own proxy.

`--host`/`--domain`/`--port` are remembered in `.rask/deploy.json` (**never secrets** — pass those via `--env`/`--env-file`).

### Design notes
- **Why the `-H ssh://` flag, not `DOCKER_HOST=`/contexts:** the `IProcessRunner` seam passes only an arg list, and every command is unit-tested by asserting the exact args. Keeping the transport in args means the hard logic lives in pure functions (`BuildRunArguments`, `BuildCaddyfile`, `NextColor`, `ParseDeployedApps`, `BuildRoutingMap`) tested directly, and the orchestration is driven by the existing `FakeProcessRunner`.
- Adds `DockerProbe` (local + remote reachability; a missing `docker` binary yields install guidance instead of crashing), `DeployConfig` (source-gen JSON), and a reusable `MultiOption` on the arg parser (repeatable `--env`).

## Testing
- 46 new unit tests; full `Rask.Cli.Tests` suite (315) green.
- Local unit gate (`scripts/run-unit-local.sh`) — all non-E2E projects green; `dotnet format` clean; warnings-as-errors build clean.
- Browser E2E gate (`scripts/run-e2e-local.sh`) — 25/25 green.
- Manually verified `rask deploy --help` and `--dry-run` (domain + port) output; the dry-run prints the exact docker command sequence.
- `rask deploy` cannot be end-to-end tested (needs a real SSH host + dockerd), so unit coverage via the process/file seam + `--dry-run` is the definition of done.

## Docs & user-facing
`docs/cli.md` (new `rask deploy` section; also drops the stale `rask info` "templates installed" row to match the CLI), `docs/deployment.md`, `docs/roadmap.md`, `docs/ai-agents.md`, `src/Rask.Cli/NUGET.md` (fixes the stale `db`/`deploy` roadmap line), `llms.txt`, and the CLI package description. CHANGELOG `[Unreleased]` → Added.

## Changelog
Added: **`rask deploy` — one-command deploy to a single host over SSH** (Docker-over-SSH build, auto-HTTPS Caddy, zero-downtime blue-green, multi-app).


pal-tamas:
Ran a high-effort automated review; the verified findings are addressed in `77eeba65`:

**Correctness / security**
- Port-mode deploy now persists `--env-file` and `--project` (they were dropped, so a bare redeploy lost env-file secrets and, in a multi-project tree, the build context).
- `--dry-run` redacts `--env` values (`KEY=…`) so previews never echo env-file secrets to stdout / CI logs.
- `--port` + `--domain` (in-flag or remembered) is now rejected with guidance — the two are distinct modes, so silently ignoring `--port` hid a 502-on-wrong-port footgun.
- Blue-green frees the target color's container name before starting it, so a leftover from a prior failed reload no longer collides on `docker run --name`.
- `caddy reload` is retried — on a fresh host the just-started proxy's admin endpoint may not be ready on the first try.

**Polish**
- Readiness/reload inspect first and wait only between retries (drops ~2s from every successful deploy).
- Tempered the "health-checked" wording (v1 gates on the container running, not an app HTTP probe — a documented follow-up).
- Added `rask deploy` to the README command tour.

New tests cover env-file/project persistence, dry-run redaction, the `--port`/`--domain` rejection, and the name-collision guard. Full `Rask.Cli.Tests` suite (320) green; browser E2E 25/25.

PR #466, opened 2026-07-17
fix(docs): stop claiming IFileSystemAccess and IWebPush work on Native

pal-tamas:
> **This PR was repurposed.** It originally "fixed" the iOS origin's secure context. That premise was refuted by a device check before merge — see [the comment below](#issuecomment-5001638876). There was no bug; the shipped code and docs were right. What survives is what the verification actually turned up.

## Summary

Three documentation claims about the browser-API surface didn't match reality. **No code changes** — nothing here was broken in code; the docs just described a surface we'd never checked.

### 1. `IFileSystemAccess` and `IWebPush` don't work on Native

Both marked ✅ in the Native column of `docs/browser-capabilities.md`. Neither API exists in the WebView — probed against the real native origin on `WKWebView`:

```
IFileSystemAccess → showOpenFilePicker = undefined     ← matrix said ✅
IWebPush          → PushManager        = undefined     ← matrix said ✅
IWebPush          → serviceWorker      = true          ← SWs register; push is what's missing
```

Both become ⬜, naming the alternative (`<input type="file">`) and the tracked APNs/FCM follow-up. **`INotifications` is unaffected** — it has a ★ native backend, so local notifications work on device.

### 2. Every wrapper's demo isn't in the section the docs point at

`docs/browser-apis.md` said every wrapper has a demo "in the **Browser APIs** section of the showcase". The *coverage* claim is true — all 46 have a runnable demo — but the **section** is wrong for 13 of them: the WASM-only tier plus `IWakeLock` and `IWebPush` have their own pages under the showcase's **PWA** group (verified: all 13 nav entries resolve to a real page). The six activation-gated ones appear in both — as gesture components under Browser APIs, as injectable services under PWA.

Pedantic-looking, but Browser APIs is where a reader is *told to look*, and a third of the surface isn't there.

### 3. Why the origins are secure contexts wasn't written down

`docs/native.md` now documents it, because only one of the two is obvious:

| Head | Origin | Secure context because |
|------|--------|------------------------|
| Android | `https://appassets.rask/` | the `https` scheme — the ordinary rule |
| iOS | `raskapp://local/` | WebKit treats a custom `WKURLSchemeHandler` scheme as trustworthy |

The iOS row goes further than the [W3C algorithm](https://w3c.github.io/webappsec-secure-contexts/#is-origin-trustworthy) alone would: `raskapp://` isn't `https` and the host isn't `localhost`, yet `isSecureContext === true`. Reading that origin as insecure and "fixing" the host is a real trap — I fell in it, which is what this PR originally was — and it matters to anyone writing a custom `INativeWebView`.

## What stops this recurring

The Native column was written from assumption. It now gets evidence: the Appium suite **prints what each WebView-only wrapper actually resolves to** on device, and **asserts the app origin is a secure context** — losing that would take `crypto.subtle`, `navigator.credentials`, `storage.estimate` and Web Locks away *silently*, as `undefined` rather than an error.

## Testing

- `dotnet format` + `dotnet build -warnaserror` clean; `scripts/run-unit-local.sh` (250/250) + `scripts/run-e2e-local.sh` pass.
- Corrections verified against `WKWebView` + `WKURLSchemeHandler` — the same shape `RaskWkWebView` uses. A negative control (`http://example.com/` → `isSecureContext=false`, `crypto.subtle=undefined`) confirms the probe discriminates rather than always reporting support.
- Verified on **macOS WebKit**, which shares WebCore with iOS. The Appium assertion + probe confirm it on a real device; the ios workload is orphaned by an SDK feature-band bump locally, so that run is still outstanding — and no longer blocking, since the probe answered the question directly.

## Follow-ups

- **`IWebAuthn` Native ✅ stands.** I expected to demote it; probing says otherwise — see [the comment below](#issuecomment-5003038659). `create()` from the native origin returns `NotAllowedError` (no focused document), **not** `SecurityError`, so origin/RP-ID validation *passes*. Still unproven: that a registration *completes* with a real authenticator on device.
- **Native push (APNs/FCM)** behind `IWebPush` — what the new ⬜ points at.

## Benchmarks

N/A — docs and an on-device test. No render hot-path or runtime code touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -184,6 +184,18 @@ them until tagged releases begin.
   README, `NUGET.md`, and the guides now targets `/docs/`.
 
 ### Fixed
+- **The capability matrix no longer claims `IFileSystemAccess` and `IWebPush` work on Native.**
+  `docs/browser-capabilities.md` marked both ✅ in the Native column, but neither API exists in the WebView:
+  `window.showOpenFilePicker` is `undefined` on WebKit (the File System Access API is effectively
+  Chromium-desktop-only), and `window.PushManager` is `undefined`, so there is nothing to subscribe with
+  (service workers do register — push specifically is missing). Both are now ⬜ with a note pointing at the
+  alternative (`<input type="file">`) and the tracked APNs/FCM follow-up. `INotifications` is unaffected — it
+  has a native backend, so local notifications work on device. No code changed: these APIs never worked on
+  Native, the matrix just said they did. The on-device Appium suite now asserts the app origin is a secure
+  context and prints what each WebView-only wrapper actually resolves to, so the Native column has evidence
+  behind it instead of assumption. `docs/native.md` also documents *why* both origins are secure contexts —
+  Android by its `https` scheme, iOS because WebKit treats a custom `WKURLSchemeHandler` scheme as
+  trustworthy whatever the host, which is easy to get wrong when writing a custom `INativeWebView`.
 - **`Rask.Outbox` now delivers nested `IOutboxEvent` types.** The source generator registers each event by
   its dot-separated display name, but `OutboxSerializerRegistry.Serialize` stored `Type.FullName` — which
   uses `+` between a nesting type and a nested type — so a nested event (a record declared inside a class)
--- a/docs/browser-apis.md
+++ b/docs/browser-apis.md
@@ -11,8 +11,11 @@ This page is the **map of the whole surface**. For an at-a-glance view of *where
 [**capability matrix**](browser-capabilities.md) — it links to a dedicated reference page per API
 under [`docs/apis/`](apis/). For the deeper "why" — user activation, the transport seam, element
 refs — see [JS interop → Typed browser APIs](js-interop.md#typed-browser-apis); for the mobile/PWA
-angle see the [Mobile & PWA guide](pwa.md). Every wrapper has a runnable demo in the **Browser APIs**
-section of the [showcase](https://pal-tamas.github.io/rask/docs/).
+angle see the [Mobile & PWA guide](pwa.md). Every wrapper has a runnable demo in the
+[showcase](https://pal-tamas.github.io/rask/docs/), under **Browser APIs** — except the WASM-only tier
+plus `IWakeLock` and `IWebPush`, which get their own pages under **PWA** because they need something the
+Server transport can't give them. (The six activation-gated ones appear in both: as gesture components
+under Browser APIs, and as injectable services under PWA.)
 
 ## Three homes, one rule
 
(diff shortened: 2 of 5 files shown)


pal-tamas:
## Withdrawing this — the premise is refuted by real WebKit

I verified the claim before merging, and it does not hold. **There is no bug.** Probing a `WKWebView` with a `WKURLSchemeHandler` — the exact shape `RaskWkWebView` uses — gives:

| Origin | `isSecureContext` | `crypto.subtle` |
|---|---|---|
| `raskapp://local/` (shipped) | **true** | `object` |
| `raskapp://localhost/` (this PR) | true | `object` |
| `http://example.com/` (control) | false | `undefined` |

The control discriminates, so the `true` is a real WebKit verdict rather than a harness artifact.

**WebKit grants a custom scheme-handler origin a secure context regardless of host.** The W3C localhost-host clause I reasoned from never gets a chance to apply here, so `local` vs `localhost` makes no difference. `crypto.subtle`, `navigator.credentials`, `navigator.locks` and `storage.estimate` were all working on iOS the whole time.

That makes every part of this PR wrong:

- The origin change fixes nothing and churns a public default.
- The class doc and `docs/native.md` claims I "corrected" were **already true**.
- The CHANGELOG entry describes a bug that doesn't exist.
- `NativeOrigins.IsPotentiallyTrustworthy` is worse than useless: it implements the W3C algorithm faithfully, but WebKit doesn't stop there, so it would report the shipped origin untrustworthy and send the next reader down the same wrong path I took.

My reasoning was sound against the spec and wrong about the engine — which is the whole reason this PR was gated on a device check rather than merged on the strength of the analysis.

Repurposing the branch for what the probe *did* find: two genuine `docs/browser-capabilities.md` errors (`IFileSystemAccess`, `IWebPush` claim Native ✅; `showOpenFilePicker` and `PushManager` are both `undefined` on WebKit), plus the on-device assertion + capability probe, which keep their value — they pin the real behaviour so the next person doesn't have to rediscover it.

pal-tamas:
## Follow-up narrowed: WebAuthn's origin validation *passes* on the custom-scheme origin

The PR notes flagged `IWebAuthn`'s Native ✅ as "API present, but is it usable?" — worth probing before anyone touches that row. Probed it, and the row is in better shape than the note implies.

WebAuthn validates the RP ID against the origin's effective domain **before** any authenticator interaction, so the error type discriminates. From `raskapp://local/`, `navigator.credentials.create()` with `rp.id` defaulted to the origin:

```
create() → NotAllowedError: The document is not focused.
isUserVerifyingPlatformAuthenticatorAvailable() → false
```

**`NotAllowedError`, not `SecurityError`.** The custom-scheme origin was *not* rejected — the call got past origin/RP-ID validation and failed only because a headless probe has no focused document or user gesture, which is what any browser would say. `isUVPlatformAuthenticatorAvailable() → false` is the absent authenticator in this process, not an origin verdict.

So the ✅ stands, and it stands for a better reason than "the API object exists". Leaving the row as-is.

**What this still does not prove:** the failure is pre-authenticator, so it shows the origin isn't rejected — not that a registration would *complete* on a real device with a real authenticator. That last step needs a focused, gestured context on device. Narrower question than the note started with, and no longer one that blocks anything here.

PR #513, opened 2026-08-04
feat(cli): scaffold every One Person Framework battery from `rask new`

pal-tamas:
> Stacked on #510 (the generator fix). Base is that branch; retarget to `main` once #510 merges.

## Summary

`rask new` documented a stack it could not actually produce. There were no flags for jobs, mail, cache, outbox, web push, snapshots or Litestream, and **`Rask.Cache` had no CLI surface at all**. Wiring any of them meant copying a page of registrations out of the docs and getting their order right by hand.

The `server` template now takes `--jobs`, `--mail`, `--cache`, `--outbox`, `--push`, `--snapshots`, `--litestream`, and `--all-batteries` for the lot. Each adds its package, its `AddRaskX<AppDbContext>()` registration, and the `modelBuilder.AddRaskX()` call that gives the pillar its tables — so `rask new Shop --all-batteries` is a running app one `rask db add Init` away.

Flags imply what they need (`--jobs` → `--data` → `--cqrs`, `--push` → `--pwa`), because every DB-backed pillar resolves `IDbContextFactory<TContext>` and push needs the service worker the PWA registration installs.

## The order is the point

The generated `Program.cs` composes the pillars in an order that is load-bearing rather than stylistic, and says why in comments:

- the **outbox before the `DbContext` factory**, so `OutboxInterceptor` is in the container when the factory callback resolves `ISaveChangesInterceptor`;
- **`ApplyRaskConventions()` after the entity configurations**, because it walks the model as it stands — run first and every entity added later silently misses the soft-delete filter and the concurrency token;
- the **Litestream restore before anything opens the database**, since restore is skipped once the file exists;
- **`AddDbContextFactory`, never `AddDbContext`** — the latter compiles and then throws at the first poll, inside a `BackgroundService`, which by default takes the host down.

`--outbox` also turns **off** the in-process domain-event publisher. Leaving it on is the nastiest trap in the set: `DomainEventInterceptor` drains and clears each entity's events before `OutboxInterceptor` can copy them, so the outbox table stays empty and delivery quietly stops being durable — while every handler still runs, so nothing looks wrong and every test still passes.

## Also

`rask generate cache <Name>` (alias `g ca`, `--feature` co-locates like `job`/`email`): a read-through accessor that owns its key and its invalidation in one place, so a stale entry is something you can find and drop rather than hunt for across inline string keys.

The twelve-boolean `GenerateServer` signature became a `ServerBatteries` record — call sites read as `batteries.Jobs` instead of a row of positional `true, false, false, true`, and a new battery no longer churns every caller and test.

## Testing

- 34 new tests: the flag implications, each pillar's package + registration + schema call, and the four ordering invariants above. Those are the valuable ones — a registration in the wrong place doesn't fail to compile and doesn't throw, it produces an app that looks fine and quietly does the wrong thing.
- Also pinned: the private VAPID key is never served by the generated endpoints; Litestream stays off until `Litestream:ReplicaUrl` is set and sets `RaskLitestreamDownload=false` so `dotnet build` never reaches GitHub releases.
- **The `--all-batteries` app is compiled under `-warnaserror`** against this commit's packed packages (`Generated_all_batteries_server_project_builds`, `RASK_CLI_BUILD_E2E=1`) — the only thing that proves a dozen composed registrations, the config-gated Litestream block, the `await` in top-level statements and the four-schema `AppDbContext` resolve together. Verified locally: passes in 22s.
- Full `scripts/run-unit-local.sh` green (686 CLI tests). Also driven by hand: `rask new Shop --all-batteries --auth --docker` and `rask generate cache PopularProducts --feature Catalog` against the real CLI.

**Benchmarks:** not applicable — no render or live-runtime code is touched.

## Docs

`docs/cli.md` (flag table, a battery section explaining the implications and the ordering, the `cache` artifact row), `llms.txt`, `CHANGELOG.md`.


pal-tamas:
Pushed a follow-up fix found by dogfooding this PR — scaffolding the showcase sample with the CLI instead of writing it by hand.

**`rask generate feature --outbox` could leave the outbox silently disabled.** The `Program.cs` splice recognised an existing registration by comparing whole first lines, so it never saw the multi-line `AddRaskData(o => { … })` that `rask new --outbox` writes, and appended a second one. `AddRaskData` is guarded so the *first* call wins — so a later call's options are dropped, and in the common flow (`rask new App --data`, then `rask g f Order --outbox`) that drops exactly the option the outbox depends on.

The consequence is the silent trap this PR's scaffold was written to avoid: `DomainEventInterceptor` drains and clears every entity's events before `OutboxInterceptor` can copy them, so the outbox table stays empty and delivery stops being durable — while every handler still runs, so nothing looks wrong.

The splice now matches on the extension method rather than the line text, and upgrades a bare `AddRaskData()` to the outbox-safe form. A hand-customised `AddRaskData` is left alone. Six regression tests in `ProgramSpliceOutboxTests`.

pal-tamas:
Rebased onto `main` after #510 merged, and **partly rewritten** — `main` moved 9 commits underneath this branch, and #511 landed continuous backup on the `--data` golden path.

Changes from what was originally written up:

- **`--litestream` is gone.** #511 already wires Litestream whenever there's a database, and does it *better* than this branch did: it derives the path from the connection string (`new SqliteConnectionStringBuilder(connectionString).DataSource`) rather than a separate `Sqlite:Path` key, which this branch got wrong — the two disagree the moment `ConnectionStrings:App` is overridden, which `rask deploy` always does. `--snapshots` now uses main's approach too.
- **The commit gating the Litestream restore is dropped** — #511 shipped the identical guard first.
- **`ProjectGenerator.Server.cs` was re-applied by hand** onto main's version rather than conflict-resolved, so #507's forwarded headers / health checks / shutdown timeout and #511's backup block are preserved by construction rather than by me resolving ten conflict regions correctly.
- **New fix carried over:** `--data` now sets `RaskLitestreamDownload=false`. Main references `Rask.SQLite.Litestream`, whose build props fetch a binary from GitHub releases unless told not to — so a scaffolded app couldn't be built offline, and errored outright on a RID with no published asset.
- **`Generated_all_batteries_server_project_builds` converted to `[SkippableFact]`** — it was using the `if (!Enabled) return;` pattern #502 deliberately removed, so it would have passed without ever running.

779 CLI tests green including the full `RASK_CLI_BUILD_E2E=1` gates; `scripts/run-unit-local.sh` green.

Issue #518, opened 2026-08-04
A decimal Input renders type=number with no step, so fractional values silently block form submit


pal-tamas:
## What happens

Bind an `Input` to a `decimal` and enter a fractional value. The field accepts the text, but submitting does **nothing at all** — no navigation, no validation message, no error, no server round-trip. Entering a whole number submits fine.

## Repro

`samples/Rask.Example.Shop`, `/orders/new` (scaffolded by `rask generate feature Order Customer:string Total:decimal`):

```csharp
Input(() => _form.Total, Id: "total")   // _form.Total is decimal
```

- Type `42` → submits, order created.
- Type `42.50` → click Save, nothing happens. The page stays on the form with both values intact.

Reproduced through Playwright against the running app: identical test, only the value differs.

## Likely cause

A `decimal` renders as `<input type="number">` with no `step` attribute. HTML's default is `step="1"`, so the browser's own constraint validation marks `42.50` as invalid and **refuses to fire the submit event** — the framework never sees it, which is why nothing surfaces at any layer.

The fix is presumably `step="any"` (or a decimal-aware step) whenever the bound type is `decimal`/`double`/`float`.

## Why it matters

This is the default output of `rask generate feature` for any money field, and money is the single most common decimal in a scaffolded app. The failure is completely silent, which makes it look like the framework's form handling is broken rather than an input attribute being wrong.

## Workaround

`Step: "any"` on the `Input`, if that's exposed.

Found while E2E-testing the Shop showcase sample. Its test currently uses a whole-number total with a comment pointing here, so the sample's outbox → mail → jobs chain stays covered.

Issue #517, opened 2026-08-04
ProjectGeneratorBuildE2ETests.Generated_wasm_hosted_solution_builds flakes on a NuGet dgspec race


pal-tamas:
## What happens

With `RASK_CLI_BUILD_E2E=1`, running the whole `tests/Rask.Cli.Tests` suite intermittently fails:

```
[auth=True,pwa=True] generated wasm-hosted solution failed to build.
NuGet.targets(198,5): error : The file '…/HE2EAP/HE2EAP.Client/obj/HE2EAP.Client.csproj.nuget.dgspec.json'
already exists. […/HE2EAP/HE2EAP.sln]
```

Run the same test **in isolation** (`--filter FullyQualifiedName~Generated_wasm_hosted`) and all four cases pass. Reproduced twice on the full run, twice green in isolation.

## Why it looks like a race, not a real build break

- The error is NuGet writing the restore dgspec, not a compile error.
- Only the `wasm-hosted` case is affected — it is the one gate that builds a **`.sln`** rather than a single `.csproj`, so the solution restore and the project restore can both write that file.
- Each test already uses its own `Guid.NewGuid()` temp directory, so it isn't cross-test collision — it's within one project's own restore.

## Impact

It makes the opt-in build gate unreliable exactly when you run it properly (all of it, before a PR), which trains people to re-run until green.

## Suggested fix

Restore once explicitly (`dotnet restore` on the `.sln`) and build with `--no-restore`, or drop the `.sln` and build the three projects directly, so only one restore writes the dgspec.

Pre-existing — not introduced by #513, which doesn't touch the wasm templates. Noticed while running the gates for that PR.

Issue #514, opened 2026-08-04
rask new pins a stale package version from a dev build, breaking restore with NU1605


pal-tamas:
## What happens

`NewCommand.LatestStableFallback` is `"0.17.0"` (`src/Rask.Cli/Commands/NewCommand.cs:29`) but the latest released version is **0.19.0**. The fallback is used whenever the running CLI's own version isn't a resolvable package — i.e. any dev or CI build, which MinVer stamps as a prerelease.

The result is a scaffolded project pinned to 0.17.0. A later `rask generate …` (or any `dotnet add package`) that resolves a newer Rask package then fails restore with a package downgrade:

```
error: NU1605: Warning As Error: Detected package downgrade: Rask.Cqrs from 0.18.0 to 0.17.0.
error:  Shop -> Rask.Data 0.18.0 -> Rask.Cqrs (>= 0.18.0)
error:  Shop -> Rask.Cqrs (>= 0.17.0)
```

## Repro

```bash
dotnet build src/Rask.Cli
mkdir /tmp/scaffold && cd /tmp/scaffold
dotnet path/to/Rask.Cli.dll new Shop --data
cd Shop && dotnet path/to/Rask.Cli.dll generate cache Report
```

## Why it matters

It only bites dev/CI builds of the CLI, not the published tool — which is exactly what makes it easy to miss, and it makes the CLI look broken to anyone building it from source or trying a nightly. The constant is also a manual step that has now been forgotten across two releases.

## Suggested fix

Either resolve the latest stable live (the `INuGetClient` the comment already anticipates), or derive the fallback from the repo's own version at build time so it cannot drift. A cheap stopgap is a unit test asserting the constant matches the newest version in `CHANGELOG.md`, so the next release trips it instead of a user.

Found while adding the battery flags in #513.

pal-tamas:
Fixed on `main` by #515 (`262df804`).

`LatestStableFallback` is gone. The pinned version is now derived from the CLI's own version rather than
hardcoded, so it can't drift from the repo again:

- a **released** CLI stamps a stable version and pins generated projects to itself;
- a **dev/CI** build stamps a MinVer prerelease that was never published, so it walks back to the release it
  came after — MinVer names a prerelease for the version it's heading *towards* (patch-bumped), so
  `0.19.1-alpha.N` came after `v0.19.0`. A minor bump (`0.20.0-alpha.N`) resolves to `0.19.0`;
[...]

PR #519, opened 2026-08-04
feat(samples): add Rask.Example.Shop, every OPF battery in one running app

pal-tamas:
> Stacked on #513 (the CLI battery flags). Retarget to `main` once that merges.

## Summary

`docs/one-person-framework.md` described an app nobody could point at. **No sample referenced `Rask.Data`, `Rask.Jobs` or `Rask.Outbox` at all** — three pillars existed only as isolated unit tests, and nothing ran an interceptor pipeline, a `JobProcessor` or an `OutboxProcessor` inside a live server with rendering and WebSocket traffic.

`samples/Rask.Example.Shop` is that app: data, CQRS, outbox, jobs, mail, cache, production SQLite, snapshots, Litestream, Web Push, PWA, auth and Docker, wired together and actually running.

## It is the CLI's output, not a lookalike

Every file was produced by `rask new Rask.Example.Shop --all-batteries --auth --docker` plus five `rask generate` commands (listed in the README). `ShopProvenanceTests` re-runs the real generators and compares them against the committed files, so "generated by the CLI" is checked rather than asserted in prose. Four files are hand-written and listed as such — the `EnsureCreated` initializer, the outbox handler body, the email content, and the Ops dashboard.

**Scaffolding it this way is what found the CLI bugs fixed in #513**, three of which were only visible by *running* the output rather than compiling it.

## The chain

Place an order and watch `/ops`: `Rask.Data` raises the event → `Rask.Outbox` commits it in the same transaction → the processor relays it via `Rask.Cqrs` → the handler queues `Rask.Mail` and schedules a delayed `Rask.Jobs` job. Every pillar's state is in the one SQLite file, which is the entire argument for DB-backed batteries.

## Testing

- **6 unit tests** over the data layer: soft delete hidden behind the global query filter but present under `IgnoreQueryFilters`, audit stamps, a real `DbUpdateConcurrencyException`, the outbox row written in the *same* `SaveChangesAsync`, and that the stored event name is one the runtime can actually resolve (a cheap tripwire for a missing analyzer reference).
- **9 Playwright tests**, each asserting a pillar *ran* rather than that a row was written — an outbox row exists whether or not delivery works. Includes reading `journal_mode`/`foreign_keys` back from the live app, which is the only real proof `UseRaskSqlite` did anything.
- Full `scripts/run-unit-local.sh` green; the Shop E2E suite passes 9/9 locally.

### Two assertions are deliberately weaker than they look

- Counts are asserted as `processed == total` rather than exact numbers: the tests share one app instance, so absolute counts would make each test depend on the order the others ran in.
- The order total is a whole number, because **a `decimal` `Input` renders `<input type="number">` with no `step`**, so the browser's own constraint validation silently blocks the submit for `42.50`. Filed as #518; the comment points there and says to restore the fractional value once fixed.

## Also found and filed

- #514 — `rask new` pins a stale package version from a dev build (NU1605 on the next `generate`).
- #517 — the pre-existing wasm-hosted build gate flakes on a NuGet dgspec race.
- #518 — the decimal input above.


pal-tamas:
Rebased onto `main`, and **the sample was regenerated** — which is the point of it being CLI output rather than a hand-written showcase.

`main` moved 9 commits underneath this branch, and #507/#511 changed what `rask new` writes. Re-running the documented command list picked all of that up automatically:

- `appsettings.json` + `appsettings.Production.json` (new from #507)
- forwarded-headers config, `AddRaskLiveSessions()`, a 15s shutdown timeout, `UseStatusCodePages()`
- continuous backup on the `--data` path, with the database path derived from the connection string
- a `Dockerfile` that copies the litestream binary

The four hand-written files the README lists (`DbInitializer`, `OrderCreatedHandler`, `OrderConfirmation`, `OpsPage`) plus the two documented `Program.cs` edits were re-applied on top. `ShopProvenanceTests` passes against the regenerated tree, so "generated by the CLI" is still checked rather than claimed.

Also note `--litestream` is gone from #513 (main already wires it on `--data`), so the sample no longer passes it.

Verified: `scripts/run-unit-local.sh` green, and the Shop browser suite 9/9 against the regenerated sample.

Issue #530, opened 2026-08-04
Element has no title attribute, so tooltips can't be expressed


pal-tamas:
### Summary

`Element` exposes `Id`, `Class`, `Style`, `Data`, `Role`, `TabIndex` and `Aria`, but not `title` — so the
global HTML `title` attribute cannot be set on any element through the typed surface.

### Why it matters

`title` is the standard way to attach a hover tooltip, and the natural use is exactly the one a data-dense
UI needs: showing a precise value behind an abbreviated one.

Concrete case from `Rask.Dashboard`: queue tables show relative timestamps ("3m ago") because that is what
an operator scans for, with the exact UTC instant meant to live in `title`. There is no way to express
that, so the absolute time had to move into the row's expandable detail instead. The same gap hits
truncated cells (`text-truncate` with the full text in `title`), abbreviations, and icon-only buttons.

### Workarounds today, and why neither is right

- `Aria: new() { ["label"] = … }` reaches screen readers but renders no visual tooltip, and overriding an
  element's accessible name to carry a timestamp is a misuse of `aria-label`.
- `Data` only emits `data-*`.

There is no general escape hatch for a plain global attribute.

### Suggested fix

Add `public string? Title { get; set; }` to `Element`, alongside the other global attributes.

**The part that needs care:** attribute render order is asserted by tests and documented in CLAUDE.md —
`id, class, style, data-*, role, tabindex, aria-*, then tag-specific`. `title` needs a defined slot in that
sequence (after `style` and before `data-*` would keep the "global attributes first" reading), and the
order tests plus any golden-HTML assertions will need updating in the same change.

Note `Style` already has a `Title` property (`src/Rask.Core/Components/Style.cs`) for `<style title="…">`;
moving the attribute onto `Element` should replace that rather than shadow it.

Issue #523, opened 2026-08-04
Battery packages ship no metrics — queue depth and dead letters aren't observable


pal-tamas:
### Summary

`docs/observability.md` promises "standard .NET primitives — no extra packages,
OpenTelemetry-exportable out of the box", but that only covers `Rask.Server`. The back-half
battery packages have **zero** instrumentation: no `Meter`, no `ActivitySource`, no health check.
Their only diagnostics are `ILogger<T>` calls.

Verified absent in `Rask.Jobs`, `Rask.Outbox`, `Rask.Mail`, `Rask.Cache`, `Rask.Data`,
`Rask.SQLite(.Litestream/.Snapshots)` and `Rask.WebPush`. The single instrumented assembly is
`src/Rask.Server/Diagnostics/` (`RaskTelemetry`, `RaskMetrics`, `RaskActivity`,
`RaskLiveHealthCheck`).

### Impact

You can export live-session and WebSocket-frame metrics, but not "how deep is the job queue",
"how many outbox events are dead-lettered", or "how long does a job take" — the numbers an
operator actually alerts on. The roadmap's *"A dead-letter surface for the background pillars"*
gap explicitly includes "no ... metric that shows you what has given up".

### Suggested instruments

Following the existing `Rask.Server` naming convention, on a meter per package (or one shared
`Rask.Jobs` / `Rask.Outbox` / `Rask.Mail` / `Rask.Cache` meter name):

- `rask.jobs.processed` / `rask.jobs.failed` — counters, tagged `job.type`
- `rask.jobs.duration` — histogram (ms), tagged `job.type`
- `rask.jobs.pending` / `rask.jobs.deadlettered` — observable gauges
- the same four shapes for outbox and mail
- `rask.cache.hits` / `rask.cache.misses` / `rask.cache.entries`

Plus an `ActivitySource` span around each job dispatch and outbox publish, so a background
failure can be correlated with the request that enqueued it.

An `IHealthCheck` reporting `Degraded` when the dead-letter count is above zero would pair well
with `AddRaskLiveSessions()`.

### Note

Gauges must be cheap — the queue-depth ones are `COUNT(*)` against SQLite, so they should be
sampled on the processor's existing poll rather than computed per observation.

Issue #522, opened 2026-08-04
Rask.Outbox has no retention or purge, so the outbox table grows forever


pal-tamas:
### Summary

`Rask.Outbox` is the only DB-backed queue pillar with **no retention or purge**, so
`OutboxMessage` grows without bound for the life of the application.

| Pillar | Retention | Purge implementation |
| --- | --- | --- |
| `Rask.Jobs` | `JobOptions.RetentionPeriod` (7d) | `JobProcessor.PurgeAsync`, hourly, paged 1000 |
| `Rask.Mail` | `MailOptions.RetentionPeriod` (7d) | `MailProcessor` `ExecuteDeleteAsync`, hourly |
| `Rask.Cache` | per-entry expiry | `CachePurger`, every 5 min |
| **`Rask.Outbox`** | **none** | **none** |

`OutboxOptions` has only `PollInterval`, `BatchSize` and `MaxAttempts` — no `RetentionPeriod`.
`OutboxProcessor.ExecuteAsync` calls `DrainAsync` only; there is no `PurgeAsync` counterpart.

### Impact

Every domain event ever raised is retained forever, with its full JSON payload. On the
single-box SQLite deployment this framework targets, an event-heavy app grows `app.db`
indefinitely — which also inflates Litestream replication and snapshot sizes. The index is
`{ ProcessedAt, Id }`, so the drain query stays fast, but the storage cost is unbounded.

### Suggested fix

Mirror the mail implementation exactly: add `OutboxOptions.RetentionPeriod` (default 7 days,
`TimeSpan.Zero` disables) and a `PurgeAsync` on `OutboxProcessor` that runs hourly and
`ExecuteDeleteAsync`es rows `WHERE ProcessedAt IS NOT NULL AND ProcessedAt < cutoff`.

Dead letters (`ProcessedAt IS NULL`) must never be purged — they are the rows an operator still
needs to see.

Issue #521, opened 2026-08-04
Jobs and Outbox processors lose a whole batch's state if one row changes concurrently


pal-tamas:
### Summary

`JobProcessor.DrainAsync` and `OutboxProcessor.DrainAsync` load a batch of tracked entities, run
every item, and then call `SaveChangesAsync` **once for the whole batch**. If any single row in
that batch is modified or deleted by anything else while the batch is being processed, EF Core
throws `DbUpdateConcurrencyException` (a tracked UPDATE/DELETE affecting 0 rows), the whole
`SaveChanges` transaction rolls back, and **none** of the batch's state is written.

`MailProcessor` does not have this problem — it saves per message.

### Impact

The side effects already happened; only the bookkeeping is lost. So on the next poll:

- **Jobs** — every job in the batch that already ran has no `ProcessedAt` and **runs again**.
- **Outbox** — every event in the batch that already published has no `ProcessedAt` and
  **publishes again**. Duplicate domain-event delivery is exactly the failure mode the outbox
  exists to prevent.

A batch is up to `BatchSize` (default 100) items, so one interfering row can cause up to 99
unnecessary re-executions per poll — and it repeats every poll until the interference stops.

### Code

`src/Rask.Jobs/JobProcessor.cs` — batch claimed at :53, mutated in the loop, single save:

```csharp
// Persist with None: jobs in this batch already ran (their side effects happened), so their
// ProcessedAt must be written even when the host is stopping — otherwise they'd re-run on restart.
await db.SaveChangesAsync(CancellationToken.None).ConfigureAwait(false);   // :101
```

The comment shows the intent is precisely that `ProcessedAt` must survive — but a batch-wide save
means one bad row defeats it for all the others.

`src/Rask.Outbox/OutboxProcessor.cs:92` is the same shape.

`src/Rask.Mail/MailProcessor.cs:110` saves inside the loop, per message, which is the correct
behaviour.

### How it gets triggered

There is no row-claiming or locking, so any concurrent writer to these tables does it:

- a manual `UPDATE`/`DELETE` against `app.db` — currently the *documented* way to deal with a
  dead letter, since `docs/roadmap.md` notes there is no UI or CLI for them;
- an operational script that prunes or re-queues rows;
- any future admin/ops surface offering retry, cancel or delete.

### Suggested fix

Save per row inside the loop, matching `MailProcessor`. That also makes the drain incrementally
[...]

PR #532, opened 2026-08-05
feat(cli): scaffold and document the batteries dashboard

pal-tamas:
**Stacked on #531 → #529 → #526.** Review those first; this branch contains their commits.

The last of the four. #529 built the dashboard, #531 gave it actions and a log tail — this makes it
reachable: scaffoldable from the CLI, mounted in the flagship sample, and documented.

## `rask new --ops`

Adds the package, the `AddRaskDashboard<AppDbContext>()` line, and — this is the part worth checking —
the authorization policy:

- **with `--auth`**, it emits a working `AddAuthorization` policy for `RaskDashboardPolicies.Access`, with
  a comment that the demo credential store has no roles so it admits any signed-in user.
- **without `--auth`**, it emits that line **commented out**, above a note that `/_ops` is Development-only
  until you add authentication.

Either way the flag cannot hand you an open dashboard. `--ops` implies `--data` (the registration has to
name a context) and is included in `--all-batteries`.

## Shop sample

`Rask.Example.Shop` now mounts the built-in dashboard **alongside** its hand-written `Features/Ops/OpsPage.cs`.

This is a deliberate departure from the plan, which said to replace the page. I kept it because the two
demonstrate different things and both are worth showing: the hand-rolled page proves *the pillars really
are just tables you can `SELECT` from*, and the built-in one proves *you don't have to write that*. Deleting
it would also have churned five passing E2E tests that use it to assert pillar behaviour, for no functional
gain. `Program.cs` says exactly this, next to the registration.

New E2E test — the fail-closed design verified end to end in a real browser: `/_ops` deep-linked
anonymously lands on `/login`; signed in, the panels for the pillars this app registered are visible.

## Docs

- New **[`docs/dashboard.md`](docs/dashboard.md)** — what each panel answers, why *failed* is the number
  that matters, the security table, the action guards, the cost of polling, and the backup probe.
- Indexed in `docs/README.md` and registered in `GuideCatalog` (the docs site has a reverse-parity guard).
- **Roadmap:** *"A dead-letter surface for the background pillars"* moves out of **Not shipped** into the
  shipped table. That entry was the original justification for this whole stack.
- **Tutorial chapter 10** keeps building the `/ops` page by hand — the exercise is the point — and gains a
  closing section pointing at the finished one, so the tutorial no longer teaches hand-rolling as the only
  option.
- `docs/cli.md` flag table, `README.md` (badge, pillar row, package table), `llms.txt`, `CHANGELOG.md`.

## Testing

- `Rask.Dashboard` added to `CliBuildE2E.FeedPackages`, so the generated project restores it from the
  commit-local feed.
- `RASK_CLI_BUILD_E2E=1 … Generated_all_batteries_server_project_builds` — **passes**. The `--all-batteries`
  scaffold now includes `--ops` with the auth policy, and it compiles.
- New `ServerBatteriesTests.Ops_implies_data_because_the_dashboard_needs_a_context`; the existing
  `Every_battery_flag_is_supported_by_the_server_template` caught the missing `TemplateCatalog` entry, which
  is exactly what it's for.
- `ShopProvenanceTests` stays green — `Program.cs` and the csproj are already on its hand-written
  exemption list, which is what let the sample take the dashboard without the generator and the sample
  disagreeing.
- `dotnet format` clean · `-warnaserror` build **0 warnings** · `scripts/run-unit-local.sh` green ·
  `scripts/run-e2e-local.sh` at baseline plus the new Shop test.

Benchmarks: not applicable — nothing on the render hot path.

## The stack

| PR | What |
| --- | --- |
| #526 | Processor durability fix + the read APIs (`LitestreamStatus`, `ListAsync`, `RecurringJobs`) |
| #529 | `Rask.Dashboard` — panels, fail-closed security, `BsStat` |
| #531 | Actions + the log tail |
| this | CLI flag, sample, docs, roadmap |

Also filed along the way: #521 (fixed by #526), #522 (outbox has no retention), #523 (no battery metrics),
#530 (`Element` has no `title` attribute).


pal-tamas:
## Verified by actually opening it

Ran the Shop sample and looked at every panel — which found two defects that all four automated gates had passed. Pushed as `651ea66f`.

**1. The documented backup probe was broken.** `docs/dashboard.md` showed `BackupProbe(LitestreamStatus litestream, …)` with a **required** dependency. `AddRaskSqliteLitestream` is config-gated on `Litestream:ReplicaUrl` in everything `rask new` scaffolds, so on any app without that setting the service isn't in the container and the documented snippet cannot be resolved.

The failure mode is the nasty kind: the probe resolves when the System panel *renders*, so the app boots clean, serves every other page, and throws `Unable to resolve service for type 'LitestreamStatus'` the first time an operator opens that one tab — in whichever environment skipped the optional config. Both dependencies are optional now, in the docs and the sample, and an absent Litestream reports `null` rather than a fabricated "stopped" (not configured ≠ not working). Two `ShopBackupProbeTests` cover it; both fail against the previous shape.

**2. The Logs panel promised a level it can't deliver.** It read *"last 500 entries at Information and above"*, but the logging pipeline applies the app's own `Logging:LogLevel` rules first — and the scaffolded `appsettings.Production.json` sets `Default: Warning`. An operator seeing no `Information` entries under that heading concludes the panel is broken. I concluded exactly that myself and spent a while chasing it. Now "at most N entries", with the interaction stated in the empty state and the docs.

The Shop sample also gains the probe, so the Backup card is exercised somewhere and the documented extension point is real code that compiles.

**Also worth recording:** the E2E fixtures with `RunPublished = true` prefer `samples/<App>/bin/Release/net10.0/publish` and **never refresh it**. My first diagnosis of #2 was wrong because I was looking at a 12-hour-old build — `run-e2e-local.sh` publishes Server/Playground/Site but not Shop.

Gate after the fixes: `dotnet format` clean · `-warnaserror` 0 warnings · **4837 unit tests, 0 failures** across 36 assemblies.

PR #559, opened 2026-08-05
feat(server): rebuild a page on a host that never knew its session

### LINKED ISSUE #548: docs/observability.md omits the reason=idle tag on rask.ws.frames.rejected
## What

`docs/observability.md:64` documents the tag values as:

> | \`rask.ws.frames.rejected\` | Counter | \`reason\` = \`size\` \| \`rate\` \| \`backlog\` | Inbound frames refused by a safety limit. |

The code emits four values, not three (`src/Rask.Server/RaskEndpointExtensions.cs`):

- `:677` `FrameRejected("size")`
- `:691` `FrameRejected("idle")` ← undocumented
- `:716` `FrameRejected("rate")`
- `:881` `FrameRejected("backlog")`

`idle` is emitted when `IdleSocketTimeout` closes a socket that has sent no inbound frame for the configured window. The XML doc on `RaskMetrics.FrameRejected` (`Diagnostics/RaskMetrics.cs:85`) does list it — only the guide is stale.

## Why it matters

`docs/observability.md:66-68` calls this counter "the headline DoS-visibility signal". Someone building an alert off the documented tag set will not have a series for `idle`, and `IdleSocketTimeout` is exactly the knob you reach for when reclaiming silently-idle connections under load.

## Fix

Add `idle` to the tag list and say what it means, alongside the existing `rate`/`backlog` explanation.

pal-tamas:
Second PR of the server-host scaling work. Stacks conceptually on **#549** (persisted data-protection key ring) but branches from `main` — no file overlap, so they can merge in either order. #549 should land first though: without a persisted key ring, every record is refused after a deploy and this degrades to exactly the behaviour it replaces.

## The problem

A live session is a component tree, a DI scope and a set of cancellation tokens. None of that serializes, so a session cannot be moved or saved — when the process holding it goes away, it is gone.

That was the end of the story. `rask deploy` replaces the container on every deploy, so every connected client got **"Your session timed out. Reload to continue."** The deploy was zero-downtime for HTTP and a full reload for everyone actually using the app.

## The approach

Nothing resumes. What travels is a small sealed record of **where the page was** and **what the app declared** through the new `IPersistentState`, and a host that receives it back **rebuilds** the page around it:

```csharp
public sealed class OrdersPage(IPersistentState state) : Component
{
    private string _filter = "";
    protected override void OnMount() => state.TryGet<string>("filter", out _filter!);
    private void Search(string term) { _filter = term; state.Persist("filter", term); }
}
```

The record is **held by the browser**, so this needs no shared store, no sticky routing and no new infrastructure — the same property that will later let a reconnect land on a different replica. Even an app that declares nothing gets the route back, which turns a deploy from a full-page reload into a re-render.

## Two things I got wrong, and what they taught

**Delivery as its own WebSocket frame.** Implemented first, then reverted — it broke ten existing tests, and they were right. The frame stream is a contract: a `hello` with nothing pending must emit **no frame at all**, and consumers reason about the last frame of a burst. The record now rides *inside* the render payload beside `history`/`auth`/`download`, which is invisible to both and semantically exact anyway — the record only changes when state or route changes, and both always come with a render. There's now a test pinning the no-frame property.

**The record inflated the diff and flipped the size heuristic.** `InitialRenderDiffTests` caught small pages switching from diff to full HTML whenever a record was due. That was a real performance defect, not a test artifact: the full-HTML payload carries the identical record, so counting it only against the diff is simply wrong. It's now discounted from the comparison.

Also found while building: **`WebApplication.CreateBuilder` does not register Data Protection** (it only appears if antiforgery/cookie-auth/session pulled it in), so `AddRask` now requests it — idempotent and additive, exactly how ASP.NET's own components do it. And **`IDataProtector.Protect` is non-deterministic**, so it cannot derive a stable user binding; that one was caught by reading before it ran.

## Security

The record is encrypted and authenticated under its own data-protection purpose. Expiry is delegated to `ITimeLimitedDataProtector`, so an expired record **cannot be opened at all** rather than relying on a field comparison. It carries **no principal** — a reconnect authenticates from its cookie or bearer token exactly as before — but is bound to a hash of the identity it was issued to and compared in fixed time, so it cannot be replayed onto another account or inherited across a sign-in. Oversized tokens are refused before decoding. A rebuild takes a `MaxSessions` slot through the **same atomic reservation a GET uses**, so the reconnect storm after a deploy sheds like ordinary traffic instead of walking past the cap.

The 16 KB bag cap degrades rather than fails: over budget the session keeps working, declares itself unresumable, and falls back to the reload it would have had, with a diagnostic saying so.

## Testing

29 new tests. `SessionHandoffProtectorTests` covers the refusals — tampered, expired, foreign key ring, wrong user, anonymous-vs-authenticated in both directions, oversized, garbage — plus **two providers sharing one key ring**, which is literally the deploy. `SessionResumeTests` drops the session from the store between two connections (what a container swap looks like from the client) and asserts the page comes back with its state and route, that a rebuilt session can itself be resumed, that refusals still reload, and that an idle session emits no frame.

Full local gate green (284 server tests, whole solution). `payload-bytes --check` **unchanged to the byte** across all five scenarios — the record costs nothing when absent. No allocation delta is possible on the null path: a virtual call and integer arithmetic, no allocation site. Browser E2E skipped (`RASK_SKIP_E2E=1`); the sandbox suite is environmentally red on `main`.

## Docs

`configuration.md` gains "surviving a restart or a redeploy" — the declare-don't-stream guidance, what does *not* come back, and the key-ring prerequisite. Plus the two new options, the two new metrics, and the note that a spike on `resume_rejected{reason=unprotect}` right after a deploy means the key ring isn't persisting. Also fixes the `reason=idle` omission (closes #548).

## Not in this PR

The drain frame (making clients reconnect immediately rather than walking the backoff ladder) — it's a latency refinement whose shutdown ordering races the per-connection `ws.Abort()`, and it belongs with `rask deploy --drain`, where the deploy-side window makes it meaningful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,33 @@ them until tagged releases begin.
   It reports no memory column on purpose: the generator shares a process with the host, so a heap reading
   counts the client's own sockets — an early version did exactly that and produced a figure that didn't
   rise with page size.
+- **A restart or a redeploy no longer costs your users their page.** A live session is a component tree, a
+  DI scope and a set of cancellation tokens — it cannot be serialized, so it cannot be moved or saved. Until
+  now that meant the process holding it going away took the page with it: every `rask deploy` blue-green
+  swap answered every connected client with *"Your session timed out. Reload to continue."* The swap was
+  zero-downtime for HTTP and a full reload for everyone actually using the app.
+  What travels instead is a small sealed record of where the page was and what the app declared through the
+  new **`IPersistentState`** — and a host that has never heard of the session **rebuilds** the page around
+  it. Nothing resumes; the page is built again, which is why what you declare is what comes back. Declare
+  the state a user would be annoyed to lose (a filter, a wizard step, an unsaved draft) and let the rest
+  come back from the database. Even an app that declares nothing gets the route, so a deploy becomes a
+  re-render rather than a reload.
+  The record is held by the browser, so this needs **no shared store, no sticky routing and no new
+  infrastructure** — the same property that will later let a reconnect land on a different replica. It is
+  encrypted and authenticated under its own data-protection purpose; expiry is enforced by ASP.NET's
+  time-limited protector, so an expired record cannot be opened at all rather than relying on a field
+  somebody remembers to check; it carries no principal but is bound to the identity it was issued to and
+  compared in fixed time, so it cannot be replayed onto another account or inherited across a sign-in; and
+  a rebuild takes a `MaxSessions` slot through the same atomic reservation a `GET` uses, so the reconnect
+  storm after a deploy sheds like ordinary traffic instead of walking past the cap. The bag is capped at
+  16 KB — over budget a session keeps working but declares itself unresumable and falls back to the reload
+  it would have had, with a diagnostic saying so.
+  It rides **inside the render payload**, beside `history` and `auth`, rather than arriving as its own
+  frame: a `hello` with nothing pending must still emit no frame at all. Costs nothing when absent — the
+  payload-byte benchmark is unchanged to the byte. Configure with `SessionResume` (default on) and
+  `ResumeTokenLifetime` (default 1 hour). **Requires a persisted data-protection key ring** — a record
+  sealed before a redeploy cannot be opened after one otherwise, which is what the scaffold's `/data/keys`
+  change below is for.
 - **`rask db backup` and `rask db restore` — get the deployed database down, and a copy back up.** Continuous
   backup (Litestream) covers the box dying; this covers the two things a solo developer actually reaches
   for: *"something looks wrong in production, let me get a copy"* and *"that migration was a mistake,
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -63,6 +63,8 @@ WebSocket safety caps and session grace periods — only the ASP.NET host has th
 | `IdleSocketTimeout` | `0` (off) | Close a connected socket that sends no inbound frame for this long (the session survives for reconnect). Reclaims silently-idle connections. |
 | `MaxPendingHandlerBytes` | `0` (off) | Aggregate-bytes companion to `MaxPendingHandlers` — bounds the queued cloned-payload *memory*, not just the queue length. |
 | `SendTimeout` | `30 s` | How long one outbound frame may take before the socket is aborted. A client that stops reading TCP would otherwise pin its session's render lock — and its disposal — indefinitely. The session survives for the grace period, so a briefly-stalled link reconnects normally. `0` disables. |
+| `SessionResume` | `true` | Let a client rebuild its page on a host that has never heard of its session — after a restart or a redeploy. See [surviving a restart](#surviving-a-restart-or-a-redeploy). Turn it off to keep the reload. |
+| `ResumeTokenLifetime` | `1 h` | How long a resume record stays redeemable. Not the reconnect grace period: that covers a blip against the *intact* session, this covers the session being gone. |
 | `HandlerTimeout` | `0` (off) | Cancel a handler's `Component.CancellationToken` after this long. A handler that threads that token into its async work unwinds cleanly instead of pinning the render pipeline (cooperative — a token-ignoring handler can't be force-aborted). |
 
 ### Reconnect UX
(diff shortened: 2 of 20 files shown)


pal-tamas:
## ⚠️ Interacts with #560 — merging both as-is makes the result worse than either alone

Cross-posting from #560 so it's visible on both. The conflict is behavioural, not textual, so `git` won't
catch it — both PRs edit `rask.js` in overlapping regions.

**This PR** rebuilds the page from the browser-held record on reconnect: no reload, state and route preserved.

**#560** (graceful shutdown drain) broadcasts a shutdown frame and closes the socket with a clean 1001, then
its client does:

```js
if (serverShuttingDown || (e && e.code === 1001)) {
    reloadForUpdate();   // → sessionExpired = true → location.reload()
    return;
}
```

It reloads **instead of reconnecting**, so `hello` never carries `resume` and this PR's rebuild never runs.
Land both as they stand and every redeploy is a full reload again, with this machinery sitting unused.

Also relevant to your "Not in this PR" note: you deferred the drain frame because its "shutdown ordering
races the per-connection `ws.Abort()`" — #560 removes exactly that race. The abort moves to the end of a
drain budget as a backstop, and the socket's token stops deriving from `ApplicationStopping` (which was why
nothing could be *sent* after SIGTERM at all). So the blocker you identified is addressed; the two were
designed to fit.

### Reconciliation

On the shutdown frame / 1001 close: show #560's "Updating…", **reconnect immediately** (skipping the backoff
ladder — the client now *knows* it's a redeploy rather than guessing), let the reconnect carry `resume`, and
fall back to a reload only on `session/unknown`, which is what this PR already does.

Net result is better than either alone: this PR removes the reload, #560 removes the wait before it.

Whichever merges second should make `reloadForUpdate` the fallback rather than the immediate action — a
small diff in one function, but a merge-order decision rather than a mechanical fixup, so I've left both
branches untouched.

pal-tamas:
## Reconciliation done and verified — branch `reconcile-560-x-559`

Rather than describe the resolution, I worked it and ran it. Pushed as
[`reconcile-560-x-559`](https://github.com/pal-tamas/rask/compare/reconcile-560-x-559) — **not a PR**, just
a worked merge of #560 × #559 so whichever lands second applies it instead of rediscovering the decisions.

**Both suites pass together: `Rask.Server.Tests` 298/298, `Rask.Core.Tests` 2038/2039 (1 skipped)** — my
drain tests and #559's `SessionResumeTests` (which drops the session from the store between two
connections, i.e. the deploy) green in the same build.

### The one judgement call

`RaskEndpointExtensions`, the socket's cancellation token:

```csharp
using var linked = CancellationTokenSource.CreateLinkedTokenSource(
    ctx.RequestAborted, drain.HardStopping);          // #560 wins over lifetime.ApplicationStopping
var resume = ctx.RequestServices.GetRequiredService<SessionResumeSupport>();
await RunSocketLoop(ws, store, limits, wsUser, resume, appFactory, linked.Token, drain.HardStopping);
```

#560's substitution has to win — it is what allows anything to be *sent* after SIGTERM, the shutdown frame
included, and it removes the `ws.Abort()` race #559 cited when deferring the drain frame. #559's `resume`
and `appFactory` arguments are kept as-is.

### Everything else is mechanical

Additive-only conflicts in `RaskServerOptions`, `RaskServerLimits`, `RaskMetrics`, `docs/observability.md`,
`CHANGELOG.md` — both sides add adjacent members; keep both. Two traps if you resolve them with a script
rather than by hand: the `RaskServerOptions` conflict straddles a doc-comment opener and a closing brace,
so a naive "keep both" produces malformed XML and an unbalanced block. Both are fixed on the branch.

**`rask.js` merges cleanly with no conflict at all**, and composes exactly as intended:

```
shutdown frame → close → #560 reconnects immediately (no reload)
                       → hello carries #559's `resume`
                       → new host rebuilds the page          ← no reload at all
                       ↓ or session/unknown
                       → #559 forgets the token
                       → #560's showSessionExpired → "Updating…" + fast reload + scroll restore
```

That is strictly better than either alone: #559 removes the reload, #560 removes the wait before it.

### One real defect the merge surfaces

`SessionResumeSupport` refuses a rebuild when `store.TryCreate` returns `null` and meters it
`resume_rejected{reason=atcapacity}`. After #560, `TryCreate` *also* returns null while the host is
draining — so a client that reconnects onto the retiring container during a swap is counted as
**at capacity** when it was actually **draining**.

The behaviour is right (that host is going away; refusing is correct). Only the tag is wrong — and it
matters more than a normal mis-tag, because #559's alerting guidance is specifically "read the
`resume_rejected` reasons after a deploy". A deploy is exactly when this fires spuriously.

Cheapest fix: check the drain flag before `TryCreate` in the resume path and tag `draining`. I have not
made that change — it is in #559's code and its author should decide whether it is a new reason value or
folded into an existing one.

Issue #550, opened 2026-08-05
Composition-guide E2E asserts 14 demos on a hub page that now has 1, so the pre-push gate is red on main


pal-tamas:
`SharedSmokeTests.TestCompositionGuideAsync` fails **deterministically on `main`** for all three
transports, so `Journey_WalksEveryPageAndUnusualActivity` is red in `ServerExampleTests`,
`WasmExampleTests` and `StandaloneWasmExampleTests` — which means the `.githooks/pre-push` E2E gate
rejects every push unless it is bypassed with `RASK_SKIP_E2E=1`.

Verified on a clean checkout of `9a040931` (no local changes).

## The assertion

`tests/Rask.Examples.E2E.Tests/SharedSmokeTests.Journey.cs:1234`

```csharp
await SideAsync("Composition", "Composition", "main .markdown-body h1");
Assert.True(await Page.Locator(".guide-demo .sample-card").CountAsync() >= 14,
    "expected the Composition guide to embed the demos (incl. component-tiers, the folded events, toast + master-detail) as live demos");
```

## Why it can never pass

The guide was split into a hub + subpages, and the demos went with the subpages. The hub the test
navigates to now embeds exactly one:

| page | `<!-- demo: -->` markers |
| --- | --- |
| `docs/composition.md` (what the test opens) | **1** |
| `docs/composition-callbacks-context.md` | 7 |
| `docs/composition-lists.md` | 10 |

So the locator counts 1 against a threshold of 14. It is a stale assertion left behind by the
hub/subpage split, not a race — it fails in ~2s, identically, every run.

Note this was previously recorded as an intermittent/sandbox-flaky test. It isn't; it is a hard
failure, and treating it as flaky is what let it sit.

## Fix

Point the assertion at where the demos actually live. Either walk the two subpages and assert their
counts separately (7 and 10, matching the markers), or keep one navigation and assert against the hub's
real content. The follow-on `#boom-throw` wait and the interactions after it also need re-pointing at
whichever page now carries them.

Worth adding a guard that ties the expected count to the markers in the `.md` files so a future docs
split fails loudly at the source rather than silently a year later.

PR #560, opened 2026-08-05
feat(server): drain live sessions on shutdown instead of severing them

pal-tamas:
## Why

`rask deploy` advertises **zero-downtime, health-gated** deploys, and the container swap really is
blue-green. But the *app* had no graceful shutdown, so what a connected user saw on every redeploy was a
frozen, blurred page and then:

> **Your session timed out. Reload to continue.**

…followed by a forced reload 4 seconds later. Nothing had timed out. The server was replaced.

The mechanism, end to end:

1. `ApplicationStopping` fired. `LiveSessionStore.CancelAllPending` cancelled the *eviction timers* only —
   it never touched the sessions.
2. `RunSocketLoop` hard-aborted every socket from a token callback (`ws.Abort()`). No close frame, so the
   browser saw code **1006**. The graceful `CloseAsync` in the `finally` was guarded on
   `ws.State == Open || CloseReceived`, and after `Abort()` the state is `Aborted` — so on the shutdown
   path it was unreachable.
3. In-flight handler chains were never awaited. A click mid-`SaveChangesAsync` was cancelled and dropped.
4. `ScheduleRemoval` degraded to a **fire-and-forget, unawaited** `RemoveAsync`, so component unmount work
   raced process exit.
5. `TryCreate` had no stopping gate, so the GET catch-all kept minting sessions that were killed on
   arrival, and `/health` stayed **Healthy** throughout shutdown.
6. The client wired `close` and `error` to one handler and **never inspected the close code**. It
   reconnected, landed on the replacement, sent `hello` with the old id, got `session/unknown`, and ran
   `showSessionExpired()` — wrong message, a wasted round trip, and a 4 s wait.

## What changed

On `SIGTERM`, in order: admission closes (`503` + `Retry-After: 1`, and a new readiness health check goes
unhealthy so a probing proxy stops routing) → every browser is told the server is going away → in-flight
handlers finish → each socket closes with a real **1001 "going away"** handshake → sessions are disposed
**awaited**, before the stop returns.

**The load-bearing change is one token substitution.** The socket's cancellation token now derives from
the drain's hard deadline instead of `ApplicationStopping`. That token becomes `LiveSession._socketCt`, so
while it was `ApplicationStopping` *every send threw the instant SIGTERM landed* — the shutdown
announcement included — and every in-flight handler was cancelled before it could finish. `ws.Abort()` is
still registered, unchanged; it just fires at the end of the budget now, where a backstop belongs.

Two details worth flagging for review, both of which invalidated my first approach:

- **`CloseOutputAsync`, not `CloseAsync`.** The drain closes from outside the receive loop, which is parked
  in `ReceiveAsync`. `CloseAsync` sends *and then receives* until the peer echoes — a second concurrent
  receive, which throws. `CloseOutputAsync` is send-only; the parked `ReceiveAsync` then observes the
  browser's echo as an ordinary close and the loop unwinds through its own `finally`.
- **Cancelling the socket token can't be the ordinary route.** Cancelling a token a `ReceiveAsync` is using
  *aborts* the WebSocket — no close frame, browser sees 1006. That is the exact behaviour being replaced,
  so "trip the token and let the loop close itself" would have reproduced the bug.

The drain also splits across two moments deliberately: hosted services stop in **reverse registration
order** and `AddRask` is typically the first line of `Program.cs`, so `StopAsync` runs *after* every
battery registered below it (a Litestream flush can spend most of the budget first). Everything that must
happen at t=0 — closing admission, flipping readiness, arming the backstop, starting the announcement —
happens in the synchronous `ApplicationStopping` callback; only the awaited parts wait for `StopAsync`.

### Client

`{"type":"shutdown","status":"draining"}` (a fixed literal, following the hot-reload frame precedent, and
deliberately **not** dev-gated). The client shows **"Updating…"**, reloads in ~250 ms instead of 4 s, and
restores scroll position and focus via `sessionStorage`. Close code `1001` is a second signal if the frame
was missed.

Form field values are deliberately **not** restored: the replacement server renders those from its own
state, and writing stale client copies over them would turn a cosmetic loss into a data one.

**A redeploy still reloads open pages, and that is not fixable here** — a live session is a component tree
plus a DI scope inside one process and cannot hand over to the next container. What this fixes is that it
now reads as an update rather than a failure.

## Config

`RaskServerOptions.ShutdownDrainTimeout`, default 5 s; `Zero` restores the old abort. Must fit inside
`HostOptions.ShutdownTimeout` — a startup warning says so when it doesn't (a warning, not a throw: a tight
ladder is suboptimal, not invalid, and refusing to start a production app over it would be the worse
failure). `rask.shutdown.sessions.abandoned` counts anything still connected when the budget ran out.

## Testing

- `ShutdownDrainTests` (10) — frame arrives before the close; close is `EndpointUnavailable`/`server-shutdown`
  and not an abort; an in-flight handler *completes* rather than being cancelled; a wedged handler costs
  the budget, not the shutdown; `503` + `Retry-After: 1` while draining; `/health` returns 503 while
  draining; sessions disposed by the time `StopAsync` returns; `Zero` restores the old behaviour.
- `ShutdownClientContractTests` (7) — source-level contract on `rask.js`: the branch returns, is *not*
  dev-gated, 1001 is inspected, the restore point is consumed before any early return and carries no form
  values, and neither the WASM nor the Native dialect carries the branch.
- `ConfigurableLimitsTests` — validation and per-host flow-through for the new option.
- Full local gate: `dotnet format` clean, `dotnet build -warnaserror` clean, **entire unit suite green**.

**One bug this caught late, worth noting:** the readiness check was first registered with `AddCheck<T>`,
which activates through `ActivatorUtilities` — public constructors only — while the check takes the
internal drain coordinator. It compiled and threw on the first real probe. Fixed with an explicit factory
registration, and `The_health_endpoint_answers_503_while_draining` now probes through the real pipeline
rather than constructing the check, which is what a direct-construction test had missed.

### Not verified

- **No benchmark delta to quote.** The suite drives `Rask.Core`'s `RenderAsLiveRoot`; nothing on that path
  changed. The one changed per-dispatch item is `LastHandlerTask` gaining volatile accessors (it has a
  second reader now — the drain — and its doc comment previously asserted single-threaded access, so the
  fence is required rather than defensive). No benchmark covers WS handler dispatch, so I have no honest
  before/after to report.
- **E2E**: 33 passed. 4 failures are all pre-existing on `main` and unrelated — three journeys fail in
  `TestCompositionGuideAsync` (open issue #550) and `PlaygroundExampleTests` is the known sandbox failure
  (no browser network for Roslyn refs). Pushed with `RASK_SKIP_E2E=1`.
- **Not run against a real deployment.** The drain is proven in-process via `TestServer`; nobody has
  watched a real container swap with this in place.

The first commit is an unrelated mechanical fix: `dotnet format --verify-no-changes` was already failing
on `main` for two Shop-sample files, so the pre-commit gate was red for everyone.

## Follow-ups (PRs 2 and 3 of this series)

- Jobs/Outbox/Mail pass the host `stoppingToken` straight into user code, so SIGTERM cancels an in-flight
  handler mid-call — for Mail that can mean a send cut *after* delivery but before the row is marked, i.e.
  a duplicate on restart. `Rask.SQLite.Litestream` already solves this correctly and is the pattern to copy.
- The budget ladder is three unrelated constants (`docker stop -t 20`, a scaffolded `ShutdownTimeout = 15s`
  coupled to it only by a comment, Litestream's 10 s) with `ServicesStopConcurrently` never set, so hosted
  services stop sequentially. Every sample except `Rask.Example.Shop` inherits .NET's 30 s default, which
  *exceeds* the docker window.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,27 @@ them until tagged releases begin.
   `ResumeTokenLifetime` (default 1 hour). **Requires a persisted data-protection key ring** — a record
   sealed before a redeploy cannot be opened after one otherwise, which is what the scaffold's `/data/keys`
   change below is for.
+- **A graceful shutdown for live sessions — a redeploy no longer tells every user their session timed
+  out.** `rask deploy` advertises zero-downtime deploys, and the container swap really is blue-green, but
+  the app had no drain: `ApplicationStopping` fired `ws.Abort()` on every socket, so the browser saw an
+  abnormal (1006) closure with no close frame, reconnected onto the replacement container, was told
+  `session/unknown`, and displayed **"Your session timed out. Reload to continue."** for four seconds
+  before reloading. Nothing had timed out. On `SIGTERM` the host now closes admission (new sessions get
+  `503` + `Retry-After: 1`, and a new readiness health check goes unhealthy so a probing proxy stops
+  routing), tells every connected browser it is going away, lets in-flight handlers finish — a click that
+  was mid-`SaveChangesAsync` used to be cancelled and dropped — closes each socket with a real `1001`
+  "going away" handshake, and disposes the sessions *awaited*, where the old path fired an unawaited
+  `RemoveAsync` that raced process exit. The client shows **"Updating…"**, reloads in ~250 ms instead of
+  4 s, and restores scroll position and focus. Form values are deliberately not restored: the new server
+  renders those from its own state, and writing stale client copies over them would turn a cosmetic loss
+  into a data one. A redeploy still reloads open pages — a session is a component tree plus a DI scope
+  inside one process and cannot hand over — but it now reads as an update rather than a failure.
+  Budget via `RaskServerOptions.ShutdownDrainTimeout` (default 5s; `Zero` restores the old abort), which
+  must fit inside `HostOptions.ShutdownTimeout`; a startup warning says so when it doesn't, and
+  `rask.shutdown.sessions.abandoned` counts anything still connected when the budget ran out.
+  The load-bearing change is a one-line token substitution: the socket's cancellation token now derives
+  from the drain's hard deadline rather than from `ApplicationStopping`, which is what makes it possible
+  to send anything at all — including the shutdown frame — after the stop signal arrives.
 - **`rask db backup` and `rask db restore` — get the deployed database down, and a copy back up.** Continuous
   backup (Litestream) covers the box dying; this covers the two things a solo developer actually reaches
   for: *"something looks wrong in production, let me get a copy"* and *"that migration was a mistake,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,16 +53,18 @@ them until tagged releases begin.
   the app had no drain: `ApplicationStopping` fired `ws.Abort()` on every socket, so the browser saw an
   abnormal (1006) closure with no close frame, reconnected onto the replacement container, was told
   `session/unknown`, and displayed **"Your session timed out. Reload to continue."** for four seconds
-  before reloading. Nothing had timed out. On `SIGTERM` the host now closes admission (new sessions get
+  before reloading. Nothing had timed out — and because the browser could not tell a deployment from a
+  crash, it had no better option. On `SIGTERM` the host now closes admission (new sessions get
   `503` + `Retry-After: 1`, and a new readiness health check goes unhealthy so a probing proxy stops
   routing), tells every connected browser it is going away, lets in-flight handlers finish — a click that
   was mid-`SaveChangesAsync` used to be cancelled and dropped — closes each socket with a real `1001`
   "going away" handshake, and disposes the sessions *awaited*, where the old path fired an unawaited
-  `RemoveAsync` that raced process exit. The client shows **"Updating…"**, reloads in ~250 ms instead of
-  4 s, and restores scroll position and focus. Form values are deliberately not restored: the new server
-  renders those from its own state, and writing stale client copies over them would turn a cosmetic loss
-  into a data one. A redeploy still reloads open pages — a session is a component tree plus a DI scope
-  inside one process and cannot hand over — but it now reads as an update rather than a failure.
+  `RemoveAsync` that raced process exit. The client shows **"Updating…"** and, because the drop is now
+  *expected* rather than guessed at, reconnects immediately instead of walking a 500 ms → 5 s backoff
+  ladder — leaving what happens to the page to the host that answers. If that host cannot rebuild the
+  session it reloads, in ~250 ms instead of 4 s, restoring scroll position and focus. Form values are
+  deliberately not restored: the new server renders those from its own state, and writing stale client
+  copies over them would turn a cosmetic loss into a data one.
   Budget via `RaskServerOptions.ShutdownDrainTimeout` (default 5s; `Zero` restores the old abort), which
   must fit inside `HostOptions.ShutdownTimeout`; a startup warning says so when it doesn't, and
   `rask.shutdown.sessions.abandoned` counts anything still connected when the budget ran out.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,8 @@ them until tagged releases begin.
   Budget via `RaskServerOptions.ShutdownDrainTimeout` (default 5s; `Zero` restores the old abort), which
   must fit inside `HostOptions.ShutdownTimeout`; a startup warning says so when it doesn't, and
   `rask.shutdown.sessions.abandoned` counts anything still connected when the budget ran out.
+  `LiveSessionStore.DisposeAsync` also became idempotent: the store is a DI singleton, so a host or a test
+  that disposed it *as well as* the container reached `Cancel()` on an already-disposed token source.
   The load-bearing change is a one-line token substitution: the socket's cancellation token now derives
   from the drain's hard deadline rather than from `ApplicationStopping`, which is what makes it possible
   to send anything at all — including the shutdown frame — after the stop signal arrives.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -76,7 +76,8 @@ wiring on your part:
 1. **Admission closes.** New sessions are refused with `503` + `Retry-After: 1`, and the `ready`-tagged
    health check goes unhealthy so a proxy or load balancer with active probes stops routing here.
 2. **Connected browsers are told.** Every live session gets a `{"type":"shutdown"}` frame, so the client
-   shows **"Updating…"** rather than the reconnect spinner.
+   shows **"Updating…"** rather than the reconnect spinner — and, because the drop is now *expected* rather
+   than guessed at, it reconnects immediately instead of walking a 500 ms → 5 s backoff ladder first.
 3. **In-flight handlers finish.** A click that is mid-`SaveChangesAsync` completes rather than being
    cancelled halfway.
 4. **Sockets close cleanly** — WebSocket status `1001` ("going away"), not an abort — and the sessions
@@ -85,13 +86,18 @@ wiring on your part:
 Anything still connected when `ShutdownDrainTimeout` elapses is aborted, and
 `rask.shutdown.sessions.abandoned` counts it.
 
-> **A redeploy does reload connected browsers, and that is not avoidable.** A live session is a component
-> tree plus a DI scope living in *that* process; the replacement container cannot inherit it. What the
-> drain fixes is the *experience*: instead of a frozen page, an abnormal disconnect, and a four-second
-> "Your session timed out", the page says "Updating…", reloads onto the new instance in ~250 ms, and
-> restores your scroll position and focus. Form field values are deliberately **not** restored — the new
-> server renders those from its own state, and overwriting them with stale client copies would turn a
-> cosmetic loss into a data one.
+> **What happens to the page itself is the replacement server's decision, not the client's.** A live
+> session is a component tree plus a DI scope living in *that* process, so the new instance cannot inherit
+> it. The client therefore reconnects and lets the host that answers decide: if it can rebuild the page,
+> nothing reloads at all; if it reports the session unknown, the client reloads — in ~250 ms, saying
+> "Updating…", and restoring your scroll position and focus.
+>
+> The drain's job is to make that drop *expected*. Before it, the socket was aborted with no close frame,
+> so the browser could not tell a deployment from a crash: it froze, backed off, and eventually announced
+> a four-second "Your session timed out" for a session that had not timed out.
+>
+> Form field values are deliberately **not** restored across a reload — the new server renders those from
+> its own state, and overwriting them with stale client copies would turn a cosmetic loss into a data one.
 
 `ShutdownDrainTimeout` must fit inside `HostOptions.ShutdownTimeout`, which must fit inside whatever your
 container runtime allows between `SIGTERM` and `SIGKILL` (`rask deploy` uses 20 s). Rask logs a warning at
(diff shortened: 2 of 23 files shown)


pal-tamas:
## Verified in a real browser

The gap I flagged in the PR body — nothing asserts what the drain *feels* like — is now closed. Ran
`samples/Rask.Example.Server` and sent it a real `SIGTERM` with a page open and scrolled, driven headlessly
via Playwright:

```
  scrollY before shutdown: 396
Sending SIGTERM to pid 62863 …

  restore point   : {"path":"/todos","y":400,"id":""}
  overlay message : "Updating…"
  shutdown frames : {"type":"shutdown","status":"draining"}
  socket events   : closed url=ws://localhost:5199/rask/ws
```

Server side, same run:

```
Rask: draining 1 live session(s) with a 5s budget.
```

…and the process exited **0** with no `rask.shutdown.sessions.abandoned` warning, i.e. the drain finished
inside its budget rather than falling through to the backstop.

Worth noting what the restore point does **not** contain: any form field values. That is the deliberate
boundary described in the PR body, now visible in the actual payload. Reasoning recorded as #571 so it
isn't relitigated from scratch later.

### Still not verified

A single-port test cannot stand up a replacement server to reload *onto*, so this proves the snapshot is
written but not that the restore lands after a real container swap. That needs the blue-green path, which
per the standing caveat has never been run against a real VPS.

pal-tamas:
## ⚠️ Interacts with #559 — merging both as-is makes the result worse than either alone

Flagging before either lands, because the conflict is behavioural rather than textual and `git` will not
catch it.

**#559** lets a host that never knew a session **rebuild** the page from a browser-held record: the client
reconnects, `hello` carries `resume`, the new host re-renders in place. No reload, state and route preserved.

**This PR** short-circuits that. `scheduleReconnect` (`rask.js`) does:

```js
if (serverShuttingDown || (e && e.code === 1001)) {
    reloadForUpdate();   // → sessionExpired = true → location.reload()
    return;
}
```

It reloads **instead of reconnecting**, so the resume token is never sent. Land both as they stand and
every redeploy is a full reload again — exactly the thing #559 exists to remove — while its machinery sits
there unused.

Worth noting the two were designed to fit: #559's "Not in this PR" section defers the drain frame because
"its shutdown ordering races the per-connection `ws.Abort()`" — which is precisely the race this PR removes.

### The reconciliation

They're complementary once the client stops choosing for the server. On the shutdown frame / a 1001 close:

1. Mark `serverGoingAway`, show **"Updating…"** (this PR's wording — correct either way).
2. **Reconnect immediately**, skipping the backoff ladder. This PR's contribution is that the client now
   *knows* a redeploy is happening, so it need not wait out 500ms→5s guessing.
3. Let the reconnect carry `resume`. The new host rebuilds → **no reload at all**.
4. Reload only where #559 already does: `session/unknown`, i.e. no record or a refused one. "Updating…"
   is still the honest wording there.

That is strictly better than either PR alone — #559 removes the reload, this PR removes the wait before it.

### What needs changing, and by whom

Whichever merges second should make `reloadForUpdate` the *fallback* rather than the immediate action. It's
a small diff in one function, but it is a **decision about merge order**, not a mechanical fixup, so I've
left both PRs as they are rather than coupling this one to an unmerged branch.

Also note the docs in this PR state plainly that "a redeploy does reload connected browsers, and that is
not avoidable" (`docs/configuration.md`, `docs/deployment.md`, CHANGELOG). #559 makes that **false**. Those
three claims need softening to "reloads unless the page can be rebuilt" as part of the same reconciliation.

pal-tamas:
## Fixed: the client now reconnects instead of reloading

Pushed `87aab5a2`, addressing the #559 interaction flagged above — and it makes this PR better standalone
too, so it isn't merely accommodation.

**What changed.** `scheduleReconnect` no longer reloads from the close handler. A known redeploy (shutdown
frame, or a 1001 close) now earns **one immediate retry**:

```js
if ((serverShuttingDown || (e && e.code === 1001)) && !shutdownRetryUsed) {
    shutdownRetryUsed = true;
    …
    attempt = 0;
    connect();          // was: reloadForUpdate()
    return;
}
```

The announcement's job is to make the drop *expected* — so the client can reconnect at once rather than
walking 500 ms → 5 s guessing whether the server is coming back. Deciding to reload from the close handler
was the client concluding the page was lost before the replacement had a say, which threw away anything the
reconnect could still recover.

- **One retry, latched.** Every failed connect closes and re-enters this handler; without
  `shutdownRetryUsed` the branch would take itself forever and hammer a host that isn't up yet. A second
  drop falls through to the ordinary ladder — the right shape when the replacement genuinely isn't ready.
- **"Updating…" is held throughout.** `updateOverlayState` now early-returns on `serverShuttingDown`, so it
  never escalates to "Still trying to reconnect…". The drop is explained and it ends.
- **`reloadForUpdate` is now the fallback**, reachable only from `showSessionExpired` — i.e. we reconnected
  and the host that answered could not carry the page over. A contract test pins that it has exactly one
  call site.

Five new client-contract tests, including that the redeploy branch contains `connect()` and **no**
`location.reload`/`reloadForUpdate`.

**Docs corrected.** The claim that a redeploy reload "is not avoidable" is now false-in-principle and was
softened in `docs/configuration.md`, `docs/deployment.md` and the CHANGELOG: what happens to the page is
the replacement server's decision, and the client's job is to ask.

### Re-verified against a real SIGTERM

```
  [ws#1] -> hello
  [ws#1] <- shutdown frame
  [ws#1] closed
  [ws#2] opened … [ws#12] opened

  sockets opened  : 12
  navigations     : 1          ← the initial load only; the page never reloaded
  last overlay    : "Updating…"
  scrollY after   : 400        ← untouched
```

Before this commit the same scenario reloaded within 250 ms. Full unit gate green.

**Not observed live:** the fallback leg (reconnect lands, host reports `session/unknown`, client reloads).
`dotnet run` takes ~40 s to bind the port locally, which is nothing like a container swap, so the
replacement never came up inside a realistic window — I stopped rather than tune the harness into
proving something artificial. That path is unchanged pre-existing logic plus a routing branch covered by
contract tests.

pal-tamas:
Two more fixes pushed (`c28f3cfd`), both self-caught.

**1. I removed an escape hatch.** The previous commit gave `updateOverlayState` a blanket early-return on
`serverShuttingDown` so a known redeploy wouldn't escalate to "Still trying to reconnect…". Right wording,
wrong scope — it also suppressed the **Retry button for the whole outage**. A deploy can fail, and a
replacement that never comes up would leave the user frozen on "Updating…" with nothing to click: strictly
worse than the reconnect state it replaced, and a regression this series introduced rather than a
pre-existing gap. The redeploy branch now keeps its own wording *and* reveals "Retry now" on the same
escalation the ordinary path uses.

**2. The drain tests were racing, and I'd fixed it on the wrong branch.** They borrowed
`GatedCounterApp.Gate` and `HangingApp.Gate` — mutable statics that `HandlerExceptionIsolationTests` and
`HandlerBackpressureTests` already own. xUnit runs those classes in parallel, so one replaces the
`TaskCompletionSource` another is parked on and both fail. I'd fixed it in #570, but the racy tests are
introduced *here* — so this PR was flaky standalone, and it may well merge first. Now uses its own
`DrainGateApp`.

It surfaced on roughly one full-suite run in four; every filtered `--filter ShutdownDrainTests` run passed.
Worth stating plainly: **a targeted run is not evidence for a test that touches shared static state.**

Gate re-run clean twice after the fix.

### Unrelated flake found along the way

Two *other* full-gate runs failed on `MailProcessorTests` and `OutboxRetentionTests` — both pass in
isolation, and neither package is touched by this branch. Timing-sensitive processor harnesses slipping
under full-suite load. Filed as #577 rather than worked around here, because the pre-commit hook turns it
into a commit blocker whose only workaround is `--no-verify` — which skips the very gate it enforces.

pal-tamas:
## Reconciliation done and verified — branch `reconcile-560-x-559`

Rather than describe the resolution, I worked it and ran it. Pushed as
[`reconcile-560-x-559`](https://github.com/pal-tamas/rask/compare/reconcile-560-x-559) — **not a PR**, just
a worked merge of #560 × #559 so whichever lands second applies it instead of rediscovering the decisions.

**Both suites pass together: `Rask.Server.Tests` 298/298, `Rask.Core.Tests` 2038/2039 (1 skipped)** — my
drain tests and #559's `SessionResumeTests` (which drops the session from the store between two
connections, i.e. the deploy) green in the same build.

### The one judgement call

`RaskEndpointExtensions`, the socket's cancellation token:

```csharp
using var linked = CancellationTokenSource.CreateLinkedTokenSource(
    ctx.RequestAborted, drain.HardStopping);          // #560 wins over lifetime.ApplicationStopping
var resume = ctx.RequestServices.GetRequiredService<SessionResumeSupport>();
await RunSocketLoop(ws, store, limits, wsUser, resume, appFactory, linked.Token, drain.HardStopping);
```

#560's substitution has to win — it is what allows anything to be *sent* after SIGTERM, the shutdown frame
included, and it removes the `ws.Abort()` race #559 cited when deferring the drain frame. #559's `resume`
and `appFactory` arguments are kept as-is.

### Everything else is mechanical

Additive-only conflicts in `RaskServerOptions`, `RaskServerLimits`, `RaskMetrics`, `docs/observability.md`,
`CHANGELOG.md` — both sides add adjacent members; keep both. Two traps if you resolve them with a script
rather than by hand: the `RaskServerOptions` conflict straddles a doc-comment opener and a closing brace,
so a naive "keep both" produces malformed XML and an unbalanced block. Both are fixed on the branch.

**`rask.js` merges cleanly with no conflict at all**, and composes exactly as intended:

```
shutdown frame → close → #560 reconnects immediately (no reload)
                       → hello carries #559's `resume`
                       → new host rebuilds the page          ← no reload at all
                       ↓ or session/unknown
                       → #559 forgets the token
                       → #560's showSessionExpired → "Updating…" + fast reload + scroll restore
```

That is strictly better than either alone: #559 removes the reload, #560 removes the wait before it.

### One real defect the merge surfaces

`SessionResumeSupport` refuses a rebuild when `store.TryCreate` returns `null` and meters it
`resume_rejected{reason=atcapacity}`. After #560, `TryCreate` *also* returns null while the host is
draining — so a client that reconnects onto the retiring container during a swap is counted as
**at capacity** when it was actually **draining**.

The behaviour is right (that host is going away; refusing is correct). Only the tag is wrong — and it
matters more than a normal mis-tag, because #559's alerting guidance is specifically "read the
`resume_rejected` reasons after a deploy". A deploy is exactly when this fires spuriously.

Cheapest fix: check the drain flag before `TryCreate` in the resume path and tag `draining`. I have not
made that change — it is in #559's code and its author should decide whether it is a new reason value or
folded into an existing one.

pal-tamas:
Merging this as the surviving implementation of the shutdown drain. #572 implemented the same feature independently (`{"type":"drain"}` via `LiveSessionDrainService` + a store-owned `Drained` token) and is being closed as superseded — this PR is the superset: admission close, a readiness health check, a real 1001 "going away" handshake, awaited session disposal, and #570's derived shutdown ladder stacked on top.

One thing #572 had that this did not, so it has been carried over as the top commit here: **`LiveSessionStore.DisposeAsync` is now idempotent**. The store is a DI singleton, so a host or a test that disposes it *as well as* the container reached `Cancel()` on an already-disposed token source. Covered by `LiveSessionStoreTests.DisposeAsync_IsIdempotent`, which exercises it with a pending removal outstanding (that removal is what owns the token source).

Gated on the integrated wave tip together with #568, #564, #559, #570 and #576: full unit gate green, browser E2E 37/38 (the one failure is `PlaygroundExampleTests`, environmentally red on `main` too — it needs browser network for Roslyn refs). The invariant worth naming: `SessionResumeTests.An_idle_session_still_emits_no_frame_at_all` still passes with this PR's broadcast frame added, so the WS frame-stream contract #559 pinned is intact.

Issue #563, opened 2026-08-05
Port-mode deploy stops the old container before starting the new one, so its health gate can never roll back


pal-tamas:
`DeployPortAsync` (`src/Rask.Cli/Commands/DeployCommand.cs:378-428`) stops and removes the old container at `:383-384` **before** starting the new one. The readiness probe then runs at `:415`, after the old container is already gone — the code comment at `:413-414` admits it "can't roll back".

So in port mode a bad image (bad config, failed migration) leaves the app down with nothing to fall back to, and the health gate is dead weight.

The downtime itself is documented and fine (`docs/cli.md:388`: "there's no zero-downtime swap on a single published port") — fixing *that* means a second port plus a proxy, i.e. becoming `--domain` mode, and isn't worth it.

But the rollback is nearly free, because the machinery already exists: `DeployPortAsync` already takes `tag`/`persist` parameters, `RollbackAsync` (`DeployCommand.Ops.cs:189-207`) already calls it with `tag: PreviousTag, persist: false`, and `PreviousImageExistsAsync` (`:238`) already checks availability. On a failed probe, when `tag == CurrentTag` and a `:previous` image exists, dump logs and re-enter `DeployPortAsync` with the previous tag — the `tag != CurrentTag` condition is its own recursion guard.

That turns "brief downtime and a broken app left running" into "brief downtime and the last known-good version back up".

Found while designing the graceful-shutdown work (#560). Note the standing caveat that `rask deploy` has still never been run against a real VPS.

PR #570, opened 2026-08-05
feat(cli): make the shutdown budget one derived ladder

pal-tamas:
PR 3 of the graceful-shutdown series. **Stacked on #560** (PR 1) — its commits are included below this
one, so review only the top commit. #566 (PR 2) is independent of both.

> Retarget this to `main` **before** #560 merges, or GitHub closes it when the base branch disappears.

## Why stacked

Because 3b below removes the stop *ordering* #560's live-session drain relies on, and I wanted to actually
run that check rather than assert it. See "The hazard" at the bottom — it holds, and is now a permanent test.

## Why

Three unrelated numbers pretending to be a budget:

- `docker stop -t 20` in `DeployCommand`
- a scaffolded `HostOptions.ShutdownTimeout = 15s`, coupled to it **only by a code comment**
- Litestream's 10s grace, plus (as of #566) mail's 10s and jobs/outbox's 5s

The generator test pinned the literal `15`, so if someone changed the deploy's 20 the test would have kept
passing while the ladder silently inverted.

Worse, `HostOptions.ServicesStopConcurrently` was never set anywhere. Stopped one at a time — the .NET
default — those inner graces **sum**: 10 + 10 + 5 + 5 = **30s against a 15s budget**. So whichever hosted
service stopped last got no grace at all, and *which one that was* depended on the order of `AddRaskX` calls
in someone's `Program.cs`. A graceful-shutdown story whose behaviour depends on the order lines appear in a
file isn't one, and it makes #566's per-pillar grace periods largely inert.

## What changed

**3a — one derived ladder.** New `ShutdownBudget` in `Rask.Cli` (which owns both `DeployCommand` and the
scaffolder). `StopTimeoutSeconds` and the scaffolded `ShutdownTimeout` now derive from it, and the tests
assert the *relationship* — app budget < deploy grace, with ≥5s margin for container teardown — instead of
the literal. The inner rungs deliberately stay non-constants: `Rask.Cli` must not reference `Rask.Jobs` /
`Rask.Mail` / `Rask.SQLite.Litestream`, so they're pinned by each package's own default test and by the
table in `docs/deployment.md`, which is now the one place all six numbers appear together. Inventing a
shared package to hold four integers would be worse.

**3b — `ServicesStopConcurrently = true`** in the scaffold and every sample, so the inner rungs overlap at
`max(...) = 10s` instead of summing to 30s.

**3c — a real pre-stop pause.** `caddy reload` returning was immediately followed by `docker stop`, with
zero time between them. Caddy hot-reload drains in-flight requests it owns, but it still holds pooled
keep-alive connections to the old upstream — a request it's about to write onto one of those when SIGTERM
lands gets a broken connection, and with the default `lb_try_duration` of 0 it is **not retried**: a 502 to a
real user. 2 seconds, sized to pool turnover, as a constant rather than a flag (it's a property of the proxy,
not of your app).

Explicitly **not** for letting live sessions migrate — a WebSocket to the old color survives until the app
closes it, so the SIGTERM *is* the migration trigger and delaying it only delays the reconnect. Draining
those is #560's job. The code comment says so, so nobody later "improves" this number for the wrong reason.

**3d — samples.** Nine of the ten web hosts inherited .NET's default **30s** `ShutdownTimeout`, which
*exceeds* the 20s `rask deploy` allows before `SIGKILL` — so a sample deployed as written would be killed
mid-shutdown. `Rask.Example.EfCore` was the sharpest case: it runs Mail and Cache and is the sample that most
looks like production. Only `Rask.Example.Shop` was right, which meant a reader copying from any other sample
inherited the wrong lesson. Guarded by a repo-scanning test, because the real failure mode is the *next*
sample somebody adds.

## The hazard, verified

#560's drain leans on hosted services stopping in reverse-registration order so it runs before Kestrel.
`ServicesStopConcurrently` removes that ordering. I said I'd verify rather than assume, so:
`The_drain_still_works_when_hosted_services_stop_concurrently` runs the full drain assertions with the flag
on. **It holds** — Kestrel's own `StopAsync` waits for in-flight requests, and a WebSocket *is* an in-flight
request, so the drain runs while Kestrel waits. That's now a permanent regression test rather than a
one-off check.

## Also fixed: a test race this series introduced

The drain tests in #560 borrowed `GatedCounterApp.Gate` and `HangingApp.Gate` — **mutable statics** that
`HandlerExceptionIsolationTests` and `HandlerBackpressureTests` already own. xUnit runs test classes in
parallel, so one class would replace the `TaskCompletionSource` the other was parked on. It surfaced as an
intermittent double failure (my drain test *and* the pre-existing reconnect test) on one full-suite run in
four. Fixed with a dedicated `DrainGateApp`; two consecutive clean full-suite runs since.

Worth flagging because it was only caught by running the whole gate repeatedly — a targeted
`--filter` run passed every time.

## Docs

`docs/deployment.md` gains **the shutdown ladder table** (the single place all six rungs appear) and states
plainly that these are budgets, not guarantees, with the metric and log line that tell you when one was
missed. `docs/cli.md`, `docs/sqlite.md` and `docs/tutorial/11-deploy.md` link to it instead of restating
numbers that will move.

One honest caveat now documented in `docs/sqlite.md`: with concurrent stop, Litestream stops alongside a job
still finishing inside *its* grace, so the last rows such a job writes may not reach the replica. Sequential
stop wouldn't reliably help either (the order is reverse-registration, so nothing guarantees Litestream is
last today), and the exposure only matters if you lose the machine in those seconds.

## Testing

`dotnet format` clean, `dotnet build -warnaserror` clean, **full unit suite green — run twice** after the
race fix. New: the ladder-relationship assertions, `ServicesStopConcurrently` in the generated `Program.cs`,
the pre-stop pause default (asserted on a plainly-constructed command, since the suite zeroes it — exactly
how a test-only default leaks into production), the samples guard, and the concurrency check above.

### Not verified

- **`rask deploy` has still never run against a real VPS** (standing caveat for this command). The pause and
  the stop arguments are asserted on built argument lists, not observed against Docker. The 502 reasoning
  is from Caddy's pooling and `lb_try_duration` default, not from a reproduction.
- **No benchmarks** — nothing here is on the render hot path.
- **E2E** unchanged and red on `main` for unrelated reasons (#550 + the Playground sandbox failure); pushed
  with `RASK_SKIP_E2E=1`.

## Deliberately not done

Port-mode's health gate still can't roll back (it stops the old container before starting the new one) —
filed as #563 rather than fixed here. Making it start-before-stop on a single published port needs a second
port plus something to swap between them, which is what `--domain` already is; reimplementing a proxy inside
the port path to avoid needing the proxy path is the wrong shape, and a shutdown PR is the wrong place to
redesign a deploy topology.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,26 @@ them until tagged releases begin.
   `ResumeTokenLifetime` (default 1 hour). **Requires a persisted data-protection key ring** — a record
   sealed before a redeploy cannot be opened after one otherwise, which is what the scaffold's `/data/keys`
   change below is for.
+- **One shutdown ladder instead of three unrelated numbers.** `docker stop -t 20` and the scaffolded
+  `HostOptions.ShutdownTimeout = 15s` were two hardcoded constants coupled only by a code comment, free to
+  drift apart silently — and the generator test pinned the literal `15`, so it would have kept passing.
+  Both now derive from a single `ShutdownBudget`, and the test asserts the *relationship* (the app budget
+  fits inside the deploy grace, with margin) rather than the number. Scaffolded apps also set
+  **`ServicesStopConcurrently = true`**, which turns out to be load-bearing rather than a tune-up: stopped
+  one at a time — the .NET default — each pillar's own shutdown grace *sums* inside the single budget
+  (10 + 10 + 5 + 5 = 30s against 15s), so whichever hosted service stopped last got no grace at all, and
+  *which* one that was depended on the order of `AddRaskX` calls in someone's `Program.cs`. Stopped
+  concurrently they overlap at 10s. `rask deploy` also now pauses briefly between pointing Caddy at the new
+  container and stopping the old one: `caddy reload` returns as soon as the config applies, but Caddy still
+  holds pooled connections to the old upstream, and a request it was about to write onto one when SIGTERM
+  landed became an un-retried 502 (`lb_try_duration` defaults to 0). There was previously no gap at all.
+
+### Fixed
+- **Every web-host sample now budgets its shutdown.** Nine of the ten inherited .NET's default 30s
+  `ShutdownTimeout`, which *exceeds* the 20s `rask deploy` allows between SIGTERM and SIGKILL — so a sample
+  deployed as written would be killed mid-shutdown. Only `Rask.Example.Shop` was right, which meant a reader
+  copying from any other sample inherited the wrong lesson. A repo-scanning test now guards the next one
+  somebody adds, since that is the real failure mode.
 - **A graceful shutdown for live sessions — a redeploy no longer tells every user their session timed
   out.** `rask deploy` advertises zero-downtime deploys, and the container swap really is blue-green, but
   the app had no drain: `ApplicationStopping` fired `ws.Abort()` on every socket, so the browser saw an
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -394,6 +394,10 @@ is waited on until its container is running **and answers an HTTP health check**
 default — the endpoint `rask new` scaffolds), then Caddy is reloaded to point at it before the old one
 is removed. If the new container fails to start, or fails its health probe, the previous version keeps
 serving. Probe a different path with `--health-path`, or skip the probe with `--no-health-check`.
+HTTP requests are zero-downtime; **live sessions re-establish**, because a session lives in the container
+being replaced and cannot hand over. The retiring container announces its shutdown first, so open pages
+show "Updating…" and reload onto the new one at their previous scroll position — see
+[the shutdown ladder](deployment.md#the-shutdown-ladder).
 
 **Multiple apps share one box.** Each app container is labelled, so the proxy's routing is regenerated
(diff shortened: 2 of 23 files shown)


pal-tamas:
Retargeted to `main`.

Two reasons, the first of which I'd missed:

1. **It was getting no CI at all.** `.github/workflows/ci.yml` triggers on `pull_request: branches: [main]`,
   so a PR based on another branch runs nothing — `gh pr checks` reported literally no checks. Benchmarks,
   commitlint and the native compile jobs had never run against this branch.
2. It removes the auto-close hazard when #560 merges and its branch disappears.

**Reviewing this PR: read only the top commit** (`86317cda`, "make the shutdown budget one derived
ladder"). The two below it are #560, which this still stacks on in content — its
`The_drain_still_works_when_hosted_services_stop_concurrently` test exercises #560's drain, which is the
whole reason it was stacked. GitHub will show the right diff once #560 merges.

Also rebased onto #560's latest (`87aab5a2`, the reconnect-instead-of-reload fix); drain tests still green
(11/11).

PR #572, opened 2026-08-05
feat(server): hand connected clients over on shutdown instead of dropping them

pal-tamas:
⚠️ **Retargeted to `main`** (was `worktree-bound-outbound-send`), so this survives #564's branch being deleted on merge.

**It still contains #564's commit, and should** — the drain broadcast runs during shutdown against a hard SIGKILL deadline, and is only safe because of that PR's bounded send and concurrent fan-out. Without them one slow client would consume the whole shutdown budget and take the SQLite checkpoint with it. So the diff below is two commits; [the drain itself is the top one](https://github.com/pal-tamas/rask/commit/22f4d8b6).

Merge order either way works, but they differ:
- **#564 first** (squash) — this branch then needs `git rebase --onto <squashed-sha> 97dcae21` and a force-push to drop the now-duplicated commit, since a squash rewrites the SHA and git won't see it as merged.
- **This one first** — it carries both changes, and #564 becomes an empty PR to close.

Completes PR 4 of the server-host scaling plan.

## The gap

A deploy swaps containers under people actively using the app. The proxy is pointed at the new container **before** the old one is stopped, so the replacement is already serving by the time anything goes away — but a browser has no way to know that. Left to itself it discovers the drop when its socket dies, then waits out a reconnect backoff (up to five seconds, page frozen) before trying an address that has been healthy the whole time.

The departing host now says goodbye: on `SIGTERM` it sends every connected client a `drain` frame **while their sockets still work**, and the client reconnects on the next tick instead of backing off. Paired with #559, a deploy stops being visible at all.

## The ordering is the whole design

Sockets are now released when the drain **finishes**, not when shutdown begins — they watch a store-owned token the drain cancels, rather than `ApplicationStopping`. `CancellationToken` callback order isn't guaranteed, so registering the aborts on the same token the drain runs from would have been a coin flip between delivering the frame and destroying the socket first. Linking the session's send token to the same thing is also what keeps it usable during the drain, which removed the need for a separate send path I'd expected to write.

## A bug my own tests caught

The drain first ran from an `ApplicationStopping` callback. Those are synchronous, so it blocked a thread on the sends — and a host shutting down under load has a busy thread pool, which is exactly when blocking on work that needs the pool stops making progress. It passed in isolation and **failed only under the full parallel suite**. It's now a hosted service: awaited by the host, running while sockets are still alive, holding no thread.

Two further test-shape bugs from the same signal, both real rather than flakes: the tests read *after* `StopAsync` returned (racing the teardown the drain precedes — a browser sits in a receive loop, so they now receive first), and they never waited for the server to actually attach the socket, so under load shutdown could legitimately drain nobody.

## Bounds

Two seconds overall, plus #564's per-send timeout. Everything after the drain — disposing sessions, the WAL checkpoint, a Litestream flush — keeps the rest of the shutdown budget, because that's the part that loses data if it runs out. A client that misses its frame simply reconnects the old way.

## No `--drain` flag

The plan called for `rask deploy --drain <seconds>`. I dropped it: the deploy already stops the old container gracefully (`docker stop -t 20`) **after** reloading the proxy, so the window exists. A flag would have added a knob that changed nothing.

## Also fixed

A double-dispose this surfaced: the store is a DI singleton, so a host or test disposing it *as well as* the container reached `Cancel()` on an already-disposed token source. `DisposeAsync` now runs once.

## Testing

Four tests: the frame arrives before the socket goes, the socket survives long enough to carry it (the ordering regression guard), shutdown isn't delayed, and an idle host doesn't trip over the drain. Full local gate green **twice consecutively** — worth stating given the first two "green" runs on this branch weren't.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,21 @@ them until tagged releases begin.
   we could neither pin nor install.
 
 ### Fixed
+- **A client that stops reading can no longer wedge its live session, or hold up delivery to everyone
+  else.** `WebSocket.SendAsync` completes when a frame reaches the transport, not when the client reads it,
+  so a client that simply stops reading TCP fills the send buffer and the send never returns. Every send
+  happens under the session's render lock, which also guards its teardown — so that one client cost its
+  session every future render *and* a `Dispose` that could never take the lock, for as long as it cared to
+  stall. A new `SendTimeout` (default 30 s, `0` disables) bounds it: on expiry the socket is aborted, which
+  unwinds the receive loop and releases the lock, while the **session** is left alone to live out its
+  normal grace period — so a briefly-stalled link reconnects to the page it already had. The default is a
+  stuck-connection backstop, not a latency budget; a slow mobile link will not trip it.
+  The store's whole-session fan-outs (`RerenderAllAsync`, `BroadcastAsync`) were also sequential, so their
+  cost was the *sum* of every session's send rather than the slowest, and one stalled client held up every
+  session behind it. Both now run with a bounded degree of concurrency. Only the dev-time hot-reload
+  signal uses them today, but this is the code the planned Broadcast pillar inherits, where a fan-out is a
+  user-facing feature — and it is a prerequisite for the deploy-drain broadcast, which has to finish
+  inside the shutdown budget before `SIGKILL` interrupts a SQLite checkpoint.
 - **A contended write no longer fails with "cannot start a transaction within a transaction".** The
   busy-retry loop re-issued the identical statement on every pass with no cleanup between attempts, and
   cleared a leaked transaction only *once*, before the loop. That caught a transaction the pooled handle
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,24 @@ them until tagged releases begin.
   (`Microsoft.Data.Sqlite`); the alternative was requiring a `sqlite3` binary on the machine, a dependency
   we could neither pin nor install.
 
+### Added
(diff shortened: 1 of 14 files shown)


pal-tamas:
Closing as **superseded by #560 + #570**, which have now merged.

Both PRs implemented the same feature independently — telling connected clients the host is going away on `SIGTERM` so they reconnect immediately instead of reading the replacement's `session/unknown` as an idle timeout. They could not both land: two frame types (`{"type":"drain"}` here vs `{"type":"shutdown","status":"draining"}` there), two hosted services, and two different owners of the socket's cancellation token.

#560 was the superset, so it won on scope rather than on quality: it also closes admission with `503` + `Retry-After`, flips a readiness health check so a probing proxy stops routing, closes each socket with a real 1001 "going away" handshake, disposes sessions *awaited*, and restores scroll and focus on the reload. #570 then stacked the derived shutdown ladder on it — including `ServicesStopConcurrently`, without which the per-pillar grace periods summed to 30s against a 15s budget.

**One thing here was uniquely yours and has been carried over**, as the top commit of #560: `LiveSessionStore.DisposeAsync` is now idempotent. The store is a DI singleton, so a host or a test that disposed it as well as the container reached `Cancel()` on an already-disposed token source. It ships with `LiveSessionStoreTests.DisposeAsync_IsIdempotent`, which exercises it with a pending removal outstanding — that removal owns the token source, so it is the case that actually failed.

The bounded send and concurrent fan-out this branch also carried are in via #564, merged separately.

Not reopening: the branch is deleted and the work is either merged or superseded.

Issue #562, opened 2026-08-05
OutboxOptions has no Validate(), so a bad value fails on the background thread instead of at registration


pal-tamas:
`AddRaskJobs` calls `options.Validate()` (`src/Rask.Jobs/RaskJobsServiceCollectionExtensions.cs:24`), and Mail and Cache do the same. **`OutboxOptions` has no `Validate()` method at all**, and `AddRaskOutbox` never calls one.

So `PollInterval = TimeSpan.Zero` doesn't fail fast at registration — it throws out of `new PeriodicTimer(...)` on the background thread in `OutboxProcessor.ExecuteAsync`, which with the default `BackgroundServiceExceptionBehavior.StopHost` takes the host down at an unrelated moment with an unrelated-looking stack.

Fix: add `internal void Validate()` covering `PollInterval`/`BatchSize`/`MaxAttempts`/`RetentionPeriod` in the same shape as `JobOptions.Validate()`, and call it from `AddRaskOutbox` right after `configure?.Invoke(options)`.

Risk is nil: every value it would newly reject already throws at runtime today, just later and less legibly.

Found while designing the graceful-shutdown work (#560); not fixed there because it is unrelated to shutdown.

PR #574, opened 2026-08-05
feat(data): PostgreSQL, SQL Server and Redis as opt-in stores, with leased claiming

### LINKED ISSUE #552: Jobs, outbox and mail processors don't lease their work — multi-instance runs everything twice
## The problem

`JobProcessor`, `OutboxProcessor` and `MailProcessor` all drain with a plain
`Where(...).OrderBy(...).Take(BatchSize)` and then run what they read. There is no claim, no lease,
and no `SKIP LOCKED`. Two instances therefore read the **same** rows and both process them:
every job runs twice, and every queued email is **sent twice**.

On SQLite this is a natural constraint and it is documented ("run one processor per app — SQLite is
single-writer"). Now that `rask new --database postgres` exists, it becomes a trap: PostgreSQL invites
`replicas > 1` / `docker compose --scale`, and nothing stops it.

## Three further hazards beyond the drain

Fixing only the drain still corrupts:

1. **Recurring jobs double-enqueue.** `JobProcessor.EnqueueDueRecurringAsync` reads `RecurringJobState`,
   sees it due, and enqueues. Two instances both do this, and the two `LastEnqueuedAt` writes race with
   no concurrency token so neither loses — N× every recurring job, forever.
2. **Jobs `PurgeAsync` uses tracked `RemoveRange` + `SaveChanges`.** Two instances purging the same page
   throw `DbUpdateConcurrencyException` and fail the whole cycle. `OutboxProcessor.PurgeAsync` already
   solved this with ids + `ExecuteDeleteAsync`, and its comment explains why; Jobs never got the same
   treatment.
3. **Mail `PurgeAsync` is one unbounded `ExecuteDelete`.** Idempotent so it is correct, but it is the
   long-write-lock shape the outbox comment warns about.

`CachePurger` is safe as-is — an idempotent set-based delete.

## Proposed fix

A portable per-row lease, claimed with `ExecuteUpdateAsync`, so one mechanism covers SQLite, PostgreSQL
and any future provider with no raw SQL:

- Two additive nullable columns per queue table: `Guid? ClaimToken`, `DateTime? ClaimedUntil`.
- Claim = read candidate ids → single `ExecuteUpdateAsync` compare-and-swap → read back what we own.
- The CAS predicate must test **lease expiry**, not `ClaimToken == null`. With the null test a row
  claimed by an instance that then dies is permanently invisible, and a sweeper becomes mandatory;
  with expiry, crash recovery falls out of the same statement for free.
- Increment `Attempts` **in the claim**, not on failure — a job that kills the process never reaches
  the failure path, so `MaxAttempts` would never dead-letter a crash loop.
- `IsConcurrencyToken()` on `ClaimToken` fences the completion write, so an instance whose lease
  expired mid-run cannot stamp `ProcessedAt` over the row another instance now owns.
- Release leases in `StopAsync`, or a rolling deploy parks a batch for the full lease duration.

Honest limit worth documenting: fencing prevents *state* corruption, not *side-effect* duplication.
Once SMTP returned, the email is out. Leases narrow the window from "always, on every instance" to
"only when an instance overruns its lease".

## Migration

The columns are nullable and additive — a plain `ALTER TABLE ADD COLUMN`, metadata-only on SQLite, no
backfill, and existing rows read as claimable. But an app that upgrades and forgets `rask db add` fails
with `no such column: j.ClaimedUntil` **inside the processor's catch**, so it logs every 5s forever while
looking healthy. Each `RunCycleAsync` catch should sniff for that and print the two commands to run.

## Not doing

- `SELECT … FOR UPDATE SKIP LOCKED` — needs raw SQL per provider, has no SQLite equivalent, and holds a
  transaction open across job execution (which makes HTTP calls).
- A leader lease (one instance drains) — scaling out would buy nothing, and a stalled leader still
  double-runs with no way to detect it.

pal-tamas:
## Summary

Opens the door `docs/one-person-framework.md` has always promised — *"when you outgrow one box, the door to a
client-server database is open"* — without moving the framework off SQLite. **SQLite stays the default and the
recommendation everywhere.**

- **PostgreSQL** (`Rask.Postgres`) and **SQL Server** (`Rask.SqlServer`) via `rask new --database`. Each is a
  drop-in for the provider's `Use…` that also applies production session settings and turns on transient-retry
  — the role `UseRaskSqlite` plays for pragmas. The provider is chosen once and **detected** from the project's
  package references thereafter, so `generate` / `db` / `deploy` cannot drift from it.
- **Jobs, mail and the outbox now lease the work they claim**, so more than one instance is safe. This was the
  blocker: those processors had no claiming at all, so two instances ran every job and *sent every email twice*.
  Natural on SQLite (single writer), a trap on a database that invites `replicas > 1`.
- **`ICache` can run on Redis**, or any `IDistributedCache`, via a non-generic `AddRaskCache()`.

Choosing a non-SQLite provider drops what treats the database as *a file* — Litestream, snapshots, the `/data`
volume, the file-copy backup. Those aren't degraded; there is no file. So `--snapshots` with a server database
is a usage **error** rather than a silently dropped flag, `rask deploy` **refuses** without a connection string
(the app would otherwise start against the placeholder compiled into `Program.cs`), and `rask db backup`
refuses and names `pg_dump`. Migrations are untouched — they forward to `dotnet ef`.

### ⚠️ Action required for existing apps

`rask db add AddLeases && rask db update`. Two nullable columns per queue table, additive, no backfill.
Skipping it is the one silent failure mode: the processors throw inside the cycle's `catch`, so the app looks
healthy while logging every five seconds and processing nothing — all three now detect exactly that and print
the two commands instead of a stack trace.

**`Attempts` changes meaning**: it counts attempts *started*, not failures (the claim increments it), so a job
that takes the process down still reaches `MaxAttempts`. Visible in `/_ops` and in metrics.

### What a lease does not do

It stops one instance overwriting another's outcome. It does not make a side effect happen once — an instance
overrunning its `LeaseDuration` can have its row taken while still working, and an email already sent is out.
At-least-once was always the contract; this narrows the window and logs which option to raise.

## Testing

- Unit: `dotnet test Rask.slnx --filter "FullyQualifiedName!~Rask.Examples.E2E"` — **green**, all assemblies.
- Build: `dotnet build Rask.slnx -warnaserror` — clean. `dotnet format` clean on every changed project.
- CLI build gate (`RASK_CLI_BUILD_E2E=1`): **5/5** — every `--database` value's scaffold actually compiles.
- Provider gate (new, `scripts/run-providers-local.sh`): **5/5** against real PostgreSQL and SQL Server —
  20 processor instances racing for 200 jobs, asserting no job is claimed twice.
- Deploy gate: passed.
- Browser E2E: **4 pre-existing failures, unrelated to this diff** — 3× `TestCompositionGuideAsync` (the
  known demo-hydration race) and 1× `PlaygroundExampleTests` (needs browser network for Roslyn refs). Verified
  by reverting this branch's only `samples/` change and reproducing the journey failure identically. Pushed
  with `RASK_SKIP_E2E=1`.

Every safety-critical test was confirmed to **fail** when its invariant is broken: the claim predicate (both
engines), the recurring-job CAS, and both pre-existing CLI defects.

## Benchmarks

n/a — nothing here touches the render hot path.

## CHANGELOG

`[Unreleased]` entries added for: PostgreSQL, SQL Server, leased claiming (with the action-required migration
and the `Attempts` semantic change), the recurring-job double-enqueue fix, the retention-sweep fixes, the
Redis-capable cache, and two pre-existing CLI defects.

## Notes for review

- **Merge order matters.** #566 and #570 touch the same three processor files (`Job`/`Outbox`/`Mail`
  `Processor` + `Options`) — this branch adds `StopAsync` overrides to all three, so whichever merges second
  needs a real resolution, not a textual one. Merges cleanly into `main` as it stands today.
- **SQL Server on arm64 runs against Azure SQL Edge**, not the full image — Microsoft's segfaults under
  emulation (exit 139). Same engine, feature subset; the gate says so rather than implying equal coverage.
  Edge is also retired by Microsoft. Full SQL Server coverage needs an amd64 host.
- Running SQL Server caught a **bad assertion in the PostgreSQL suite** from the earlier phase: it required all
  200 jobs to be claimed in one round, which passed on PostgreSQL by timing alone. Fixed in both.
- Deviated from the plan in one place: `rask db backup`'s refusal is a guard rather than an extracted
  `IDatabaseBackupStrategy`. The interface would have had exactly one implementation. Same behaviour, less
  ceremony — happy to add the seam if you'd rather.
- Closes #552.
- Happy to split this into four PRs along the phase boundaries if you'd prefer to review it that way.


pal-tamas:
Merged as part of the merge-train after #566 (shutdown grace for the three processors), which required one behavioural reconciliation worth recording.

**A shutdown must not count an attempt against leased work.** This PR moves `Attempts` to count attempts *started* — the claim increments it — so an item that takes the process down still reaches `MaxAttempts`. #566 independently gives the in-flight item a bounded grace on `SIGTERM`. Where the two meet, the naive merge (which git produces cleanly, with no conflict in any of the three processors) makes `MaxAttempts` a function of **deploy cadence**: at the default of 10, ten unlucky redeploys dead-letter an item that never once failed, and nothing in the logs connects the two.

The roll-back lives in each processor's `StopAsync`, next to the existing lease hand-back, rather than in the shutdown `catch`. Two reasons:
- the `catch` ends in `break`, which skips the per-item `SaveChangesAsync`, so an in-memory `Attempts--` there is silently discarded while the claim's increment is already on disk (confirmed — the first attempt failed exactly this way);
- `StopAsync`'s `ClaimToken == _lease && ProcessedAt == null` predicate also covers items claimed in the batch but **never dispatched**, which carry the same unearned increment. An item that genuinely failed released its own lease, so it is correctly excluded.

Also reconciled: `OutboxOptions.Validate()` (both PRs added one — merged to the union of all checks, which incidentally closes #562), and `ShutdownBudgetBlock` now takes a `fileBasedDatabase` flag so a Postgres scaffold isn't told to size its shutdown budget around a WAL checkpoint and a Litestream flush that don't exist for it.

Gates on the integrated tip: unit green, CLI build gate 26/26 (every `--database` scaffold compiles), and the **provider gate 5/5 against real PostgreSQL and SQL Server** — 20 concurrent claims never double-assign, expired leases reclaim exactly once.

Issue #646, opened 2026-08-08
rask generate job prints server-only next steps inside a WASM app


pal-tamas:
## What happens

`JobGenerator.Notes` (`src/Rask.Cli/Scaffolding/JobGenerator.cs:42-55`) prints the same next steps regardless of what kind of app you are standing in:

```
  1. Register the services in Program.cs (once):
       builder.Services.AddRaskCqrs();
       builder.Services.AddRaskJobs<AppDbContext>();   // your DbContext
       builder.Services.AddDbContextFactory<AppDbContext>(o => o.UseSqlite("Data Source=app.db"));
  ...
  3. Create the schema:
       rask db add AddJobs && rask db update
```

Run `rask g j` in a browser/WASM app and every one of those lines is wrong for it:

- `Data Source=app.db` is a path in the WASM runtime's in-memory filesystem. It works, and then silently vanishes on reload, because nothing persists it.
- `rask db add` / `rask db update` wrap `dotnet-ef` against a design-time database. There is no migrations assembly in a browser bundle, so this is not a step the user can take.
- The registration omits the one call that is actually required in a browser — `AddRaskBrowserSqlite(...)` — without which the queue is not durable at all.

So the scaffolder emits instructions that cannot be followed, and the failure mode is not an error: the app builds, runs, and quietly loses every queued job on reload.

## Why it is reachable

`rask new --wasm` does not offer a jobs battery (`TemplateCatalog.cs` gives the `wasm` template only `["auth","pwa","docker"]`), but `rask generate job` is not gated on project kind — it runs anywhere a `.csproj` is found.

## Fix

`ProjectContext` already detects the database provider by reading the project file. Browser-ness can be detected the same way (`net10.0-browser` TFM / `RaskWasm` / a `Rask.Wasm` reference) and `Notes` branched on it, so a WASM app is told to register `AddRaskBrowserSqlite` and to create the schema at boot rather than through `rask db`.

Found while adding browser-side job support (`Rask.SQLite.Browser` + hosted services on the WASM host).

pal-tamas:
Both halves are on main.

- **#648** branched `JobGenerator.Notes` on project kind, so a browser app is told to register `AddRaskBrowserSqlite` and create its schema at boot instead of being sent to `rask db` — and made browser jobs actually work, which is why the notes have something true to say.
[...]

Issue #642, opened 2026-08-08
epic: backendless multi-client sync — SQLite in the browser, per-client op-logs in any object store


pal-tamas:
> **Rewritten 2026-08-08.** The original scoped this as one-way hydration of a read-only replica from a Litestream replica. The actual goal is **multi-client, offline-and-online, no backend at all**, which needs a different design. The Litestream investigation and its findings are preserved at the bottom.

## Goal

Let a Rask WASM app be a **complete application with no backend** — multiple clients, working offline, syncing through nothing but an object storage bucket. The app deploys to any static host (Azure Storage static website, S3, GitHub Pages, Netlify). There is no server process anywhere: no API, no sync service, no database tier.

The local store is SQLite in the browser, queried with EF Core, so application code is LINQ against a real database rather than a bespoke sync API.

## Design: per-client append-only op-logs

Every client writes **only to its own prefix**: `clients/<clientId>/ops/<seq>`. Append-only, never mutated.

This is the design decision everything else follows from. Object storage has exactly one strength — single-writer-per-key with cheap prefix listing — and this uses only that. **No client ever writes a key another client writes, so there are no write conflicts at the storage layer at all.** No lease, no lock, no contention, no retry-on-conflict. Offline falls out for free: queue ops locally, flush the prefix on reconnect.

Each client lists the prefixes, pulls ops it hasn't seen, and applies them locally.

**The SQLite database is a materialized view; the op-log is the source of truth.** That inversion is what makes the system recoverable — a corrupted or lost local database is rebuilt by replaying, not restored from a backup. It's also what makes a new device trivial: it has no state to reconcile, it just replays.

### Merge needs exactly two things

**A clock that survives multiple devices.** Wall clocks drift and run backwards, so ops cannot be ordered by `DateTime.UtcNow`. A hybrid logical clock — physical ms, logical counter, client id — is small and gives a total order that respects causality.

**A conflict rule.** Per-field last-writer-wins is the pragmatic default. Deletes must be **tombstones**: actually removing a row resurrects it the moment an older op from another client arrives.

LWW **loses data by design** — two clients edit the same field offline and one edit disappears. This is the single most important UX problem in the epic, not a footnote. Conflicts must be *reported*, not silently resolved. A user who loses work to an invisible merge is the failure mode that would define this feature.

### Compaction

[...]

PR #648, opened 2026-08-08
feat(wasm): run background jobs in the browser with the server's syntax

pal-tamas:
## Summary

Background jobs now run in a browser WASM app **with the server's syntax** — `AddRaskJobs<AppDbContext>()` is
literally the same line on both hosts, and a job plus its `ICommandHandler` compiles and runs on either.

`Rask.Jobs` itself needed **zero changes**. `JobQueue<TContext>` and `JobProcessor<TContext>` are already
EF-generic, so the work was entirely in the host: give the browser a way to start hosted services, and give it
a SQLite database that survives a reload.

```csharp
host.Services.AddRaskBrowserSqlite("app");                       // the only browser-specific line
host.Services.AddDbContextFactory<AppDbContext>(o => o.UseSqlite(BrowserSqlite.ConnectionString("app")));
host.Services.AddRaskCqrs();
host.Services.AddRaskJobs<AppDbContext>();
```

Five commits, each independently reviewable:

1. **`feat(wasm)` — run registered hosted services in the browser host.** `AddHostedService` always *compiled*
   on WASM but nothing ever *started* it, so a `BackgroundService` registered fine, resolved fine, and silently
   never ran. Started at the end of boot; drained from `pagehide` (skipping bfcache suspends).
2. **`feat(core)` — store raw bytes in IndexedDB.** `SetBytesAsync`/`GetBytesAsync` as default interface methods:
   base64 on the wire, a real `Uint8Array` in storage, so a database costs its own size in quota rather than a
   third more.
3. **`feat(sqlite)` — `Rask.SQLite.Browser`.** Restore from IndexedDB in `StartAsync`, snapshot back via the
   Online Backup API, one owner tab via a Web Lock.
4. **`feat(samples)` — the browser-jobs sample + E2E.**
5. **`fix(cli)` — `rask generate job` in a browser app (#646).**

## Verification

Verified in a real headless browser against a `net10.0-browser` publish: job queued → `JobProcessor` claims it
with its lease → handler writes a row → page repaints out-of-band from the poll loop → snapshot to IndexedDB →
reload → database restored (28,672 bytes) → **the greeting survives**.

- `dotnet format` clean; `dotnet build Rask.slnx -warnaserror` clean.
- Full unit suite green (pre-commit gate ran on every commit). New: 14 in `Rask.Wasm.Tests`, 40 in
  `Rask.SQLite.Browser.Tests`, 6 in `Rask.Core.Tests`, 10 in `Rask.Cli.Tests`.
- New E2E: `BrowserJobsWasmExampleTests` — queue a job, assert the row the processor wrote, reload, assert it
  survived. Run against a real published bundle in headless Chromium: **passing**.
- No benchmarks: nothing here touches the render hot path.

## Three constraints, stated rather than hidden

- **EF Core does not survive the trimmer in a browser build** (`MissingMethodException` on
  `EntryCurrentValueComparer<int>`). An app using it must publish with `PublishTrimmed=false`. Raw
  `Microsoft.Data.Sqlite` survives both trimming and AOT. This is why the sample is its own project rather than
  a page in the showcase, whose `TrimMode=full` zero-IL-warning gate stays intact.
- **The durability window is the snapshot interval, not the page-hide flush.** The browser does not wait for a
  `pagehide` handler, so a force-closed tab loses whatever changed since the last tick.
- **One tab owns the database.** The others get their own empty, unpersisted database and log why. Promotion
  when the owner closes, and proxying writes to the owner, are not implemented.

## Cost to be aware of

`scripts/run-e2e-local.sh` gains one publish that **must not** use `-p:WasmBuildNative=false` — SQLite is a
native library, and skipping the emscripten relink yields a bundle that boots and then fails on every database
call. That is a multi-minute relink. The fixture inspects the published output and says so if it was built the
wrong way.

## Notes

- The native link needs no hand-wiring: the SQLitePCLRaw 3.x bundle already pinned for CVE-2025-6965 resolves to
  a native package that ships `runtimes/browser-wasm/nativeassets/net9.0/e_sqlite3.a` and adds the
  `NativeFileReference` itself when `RuntimeIdentifier=browser-wasm`.
- `docs/sqlite.md`'s OPFS paragraph was scoped in conversation with the #642 session: OPFS makes the *flush*
  incremental, it does **not** move the live database out of MEMFS — that needs `createSyncAccessHandle`, which
  exists only inside a Worker while Rask boots the runtime on the main thread.
- `Rask.SQLite.Snapshots` gains an `InternalsVisibleTo` for the browser package, so the Online Backup logic
  stays in one place instead of being reimplemented; the shared hosted service snapshots unconditionally, and in
  a browser that has to be gated on tab ownership.


pal-tamas:
Closing #651 in favour of this — I'd independently taken #646 and landed on refusing `rask generate job` in a browser app, on the grounds that `AddRaskJobs` registers a hosted service the WASM host never starts. This PR removes that premise by starting them, so branching the notes is the right answer and a refusal would be wrong. Handing over the one thing from that branch worth keeping.

**`DetectBrowser` false-positives on the hosted-WASM Server project**, which is the case that matters most here: it's where jobs *do* belong and where the server notes are correct.

```csharp
return csprojText.Contains("-browser", StringComparison.OrdinalIgnoreCase)
    || csprojText.Contains("<RaskWasm>", StringComparison.OrdinalIgnoreCase)
    || csprojText.Contains("Rask.Wasm", StringComparison.OrdinalIgnoreCase);
```

`ProjectGenerator.WasmHosted.cs:416` gives the **Server** project `<PackageReference Include="Rask.Wasm.Hosting" …>`, which contains `Rask.Wasm` — so `rask g j` in the Server half of a `wasm-hosted` solution prints the browser steps and adds `Rask.SQLite.Browser` to a project that should be told to use `rask db`. The PR body's "cost of a false positive is next-steps text mentioning a package the reader doesn't have" holds for a stray match, but not for this one: it's a whole scaffolded template, and it's the project the browser notes send you *to*.

Three tightenings, all narrowing:

```csharp
// A browser TFM on either element — matched by the "-browser" suffix rather than the version, so a
// framework bump doesn't silently stop detecting it.
[GeneratedRegex(@"<TargetFrameworks?>[^<]*-browser", RegexOptions.IgnoreCase)]
private static partial Regex BrowserTargetFrameworkRegex();

return BrowserTargetFrameworkRegex().IsMatch(csprojText)
    || csprojText.Contains("<RaskWasm>true</RaskWasm>", StringComparison.OrdinalIgnoreCase)
    || csprojText.Contains("Include=\"Rask.Wasm\"", StringComparison.OrdinalIgnoreCase);
```

1. **Scope `-browser` to the TargetFramework element.** A bare `Contains` matches the string anywhere — a comment, a constant, a package id.
2. **Check `<RaskWasm>true</RaskWasm>`, not `<RaskWasm>`.** The bare form also matches `<RaskWasm>false</RaskWasm>`, which asserts the opposite.
3. **Keep the closing quote on `Include="Rask.Wasm"`.** That's what excludes `Rask.Wasm.Hosting`.

And the test that pins it:

```csharp
[Fact]
public async Task Job_in_the_server_half_of_a_hosted_wasm_solution_is_still_scaffolded()
{
    // Rask.Wasm.Hosting is referenced by the SERVER project, and it must not read as "browser" —
    // that project is exactly where the job belongs.
    var console = new StringConsole();
    var fs = new FakeFileSystem();
    fs.Seed(
        "/proj/MyApp.csproj",
        """<Project><ItemGroup><PackageReference Include="Rask.Wasm.Hosting" Version="1.0.0"/></ItemGroup></Project>""");
    var command = new GenerateCommand(console, fs, new FakeProcessRunner(), ProjectDir);

    var exit = await command.ExecuteAsync(["job", "SendWelcomeEmail"], CancellationToken.None);

    Assert.Equal(0, exit);
    Assert.True(fs.Files.ContainsKey(Path.GetFullPath("/proj/Features/Shared/SendWelcomeEmail.cs")));
}
```

Worth asserting the notes text on that one too, not just the exit code — the failure here is that it scaffolds *successfully* with the wrong instructions.

PR #651, opened 2026-08-08
fix(cli): refuse to scaffold a background job into an app that can never run it

### LINKED ISSUE #646: rask generate job prints server-only next steps inside a WASM app

pal-tamas:
## Summary

`rask generate job` wasn't gated on project kind, so it ran inside a WASM app and printed next steps
written for a server — `AddRaskJobs<AppDbContext>()`, `Data Source=app.db`, then
`rask db add && rask db update`. Every line is wrong there: `rask db` drives `dotnet-ef` against a
design-time database a browser bundle doesn't have, and the connection string names a file in the WASM
runtime's **in-memory** filesystem that disappears on reload.

## I diverged from the issue's proposed fix, and here's why

#646 proposed branching `Notes` on browser-ness and describing the browser wiring. Two things it didn't
have:

1. **There is no working shape to point at.** `AddRaskJobs` registers the processor as a **hosted
   service** (`RaskJobsServiceCollectionExtensions.cs:32` → `JobProcessor<T> : BackgroundService`), and
   `WasmHostBuilder.RunAsync` builds the provider and **never starts hosted services**. So the queue has
   no runner at all in a browser app — a deeper failure than the persistence one the issue describes.
2. **`AddRaskBrowserSqlite` doesn't exist on main.** It's part of the in-flight #642 epic. Writing next
   steps around it would ship instructions for an API that isn't there — the same defect class the issue
   objects to, just aimed at the future.

So rather than reword the notes, the CLI **refuses**. That was checked with you before implementing.

## What changed

- `ProjectContext` learns `IsBrowser`, read off the project file the way `Provider` already is: a
  `-browser` target framework, `<RaskWasm>`, or a `Rask.Wasm` package reference. The package check keeps
  its closing quote, so **`Rask.Wasm.Hosting` deliberately doesn't match** — that's referenced by the
  *server* half of a hosted-WASM solution, which is exactly where the job belongs.
- `generate job` stops in a browser project and says why, and where to put it:

```
Background jobs don't run in a browser app, so 'rask generate job' won't scaffold one here.
  AddRaskJobs registers the processor as a hosted service, and the WASM host never starts
  those — the job would compile and enqueue, and nothing would ever run it. The queue also
  needs a database that survives a reload, which the browser's in-memory file isn't.
  Generate it in a server project instead (in a hosted-WASM solution, the Server one):
      rask generate job SendWelcomeEmail --project <path-to-server-project>
```

- **Exits 1, not the usage code.** The command line was fine; `rask generate --help` has nothing to say
  about this, so printing usage would send you looking in the wrong place. This matches the existing
  "no project found" branch rather than the argument-validation ones.
- Every other generator still runs in a WASM app.

## Testing

- 6 new cases in `GenerateCommandTests`: the refusal via each of the three detection signals (theory),
  the `rask g j` **alias** (the check keys off the resolved verb, and the alias is how people actually
  type it), `Rask.Wasm.Hosting` still scaffolding, and `generate page` still working in a WASM app.
  They assert the exit code, that **nothing was written**, that the way out is in the message, and that
  no usage block is printed.
- **Verified on the real CLI**, not just the fake filesystem: `rask new -t wasm` → `rask g j` exits 1 and
  writes nothing; the same command in a scaffolded server app exits 0 and writes
  `Features/Shared/SendWelcomeEmail.cs` with its packages added.
- `dotnet format` clean, `CI=true dotnet build Rask.slnx -warnaserror -m:1` clean, full unit suite green,
  browser E2E 57/57.
- Docs: `docs/cli.md` marks the verb refused in WASM and explains it under the generate table.

Closes #646.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,23 @@ them until tagged releases begin.
 
 ## [Unreleased]
 
+### Fixed
+- **`rask generate job` no longer scaffolds a job into a browser app that could never run it.** The
+  generator wasn't gated on project kind, so running it inside a WASM app produced an `IJob` + handler
+  and printed next steps written for a server: register `AddRaskJobs<AppDbContext>()`, point EF at
+  `Data Source=app.db`, then `rask db add && rask db update`. Every line of that is wrong there —
+  `rask db` drives `dotnet-ef` against a design-time database a browser bundle doesn't have, and the
+  connection string names a file in the WASM runtime's **in-memory** filesystem that disappears on
+  reload. Worse than either: `AddRaskJobs` registers the processor as a **hosted service**, and
+  `WasmHostBuilder.RunAsync` never starts those — so the queue would have had no runner at all. The
+  scaffolded code compiled, enqueued, and silently did nothing.
+  The CLI now recognises a browser project (a `-browser` target framework, `<RaskWasm>`, or a
+  `Rask.Wasm` reference) and refuses, saying why and where the job belongs — in a hosted-WASM solution,
+  the Server half: `rask generate job SendWelcomeEmail --project ../MyApp.Server`. It exits 1 rather
+  than as a usage error, because the command line was fine; `--help` has nothing to say about it. Every
+  other generator still runs in a WASM app, and `Rask.Wasm.Hosting` (referenced by the *server* half of
+  a hosted solution) deliberately doesn't read as "browser".
+
 ### Added
 - **`Mount` — give a component you built yourself the lifecycle it was missing.** A component normally
   enters the tree through its generated factory, and that factory is what registers the instance with its
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -206,7 +206,7 @@ folder path, the C# convention), and **refuses to overwrite an existing file** u
 |----------|-------|-------------------|
 | `page <Name>` | `Features/<Name>/<Name>Page.cs` — a routed page `Component` with a `Head` title | `<Name>Page` in `<Root>.Features.<Name>` |
 | `component <Name>` | `Features/Shared/<Name>.cs` — a plain `Component` (or `Features/<Feature>/` with `--feature`) | `<Name>` in `<Root>.Features.Shared` |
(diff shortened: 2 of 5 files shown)


pal-tamas:
Superseded by #648, which fixes #646 by making browser jobs actually work (hosted services in the WASM host + Rask.SQLite.Browser) rather than refusing to scaffold them. That removes the premise this PR rested on — that AddRaskJobs registers a hosted service the WASM host never starts — so the refusal here would be wrong once #648 lands, and the two conflict in ProjectContext.cs besides. The detection hardening and the hosted-WASM-Server regression test are handed over in https://github.com/pal-tamas/rask/pull/648#issuecomment-5225290059.

Issue #697, opened 2026-08-17
RASK014 and RASK023 quick-fixes still emit the factory, while their analyzers now say "chain"


pal-tamas:
Found while bringing the docs onto the chain (#696).

Two shipped IDE quick-fixes still speak — and emit — the factory, while the analyzers they fix have already moved to the chain. A user hitting the error reads one thing and the lightbulb writes another.

### RASK014

The descriptor (`ComponentConstructionAnalyzer`) now reads:

> **Components must be built through a chain**
> Do not instantiate 'X' with 'new'; name it and chain onto it instead — 'X.Prop(value)', or 'X' alone when it needs nothing

But `ComponentConstructionCodeFixProvider` is titled **"Use the generated factory"** and rewrites `new Widget()` into `Widget()` — a factory *invocation*.

For the argument-free construction it is willing to fix, the chain spelling is the **bare entry** `Widget`, with no parentheses (an `IPropertyReferenceOperation`, not an invocation). So the fix should produce `Widget`, not `Widget()`.

It still compiles today only because the factory is still emitted; it breaks the moment the factory is dropped, and meanwhile it teaches the old spelling at exactly the moment a migrant first hits the error — RASK014 is an **Error**, and it is the first thing a Blazor/plain-C# migrant trips on.

### RASK023

`ImgMissingAltCodeFixProvider` is titled **`Add Alt: "" (decorative image)`** and inserts `Alt: ""`, a named argument. On a chain (`Img.Src("/x.png")`) the fix should append `.Alt("")`.

### Also

`docs/diagnostics.md`'s quick-fix table still describes both in factory terms. That is currently *accurate* (it documents what the fixes really do), so it should be updated in the same change rather than before it — #696 deliberately left those two rows alone for that reason.

### Suggested

- Retitle both fixes and emit chain syntax (`Widget` bare entry; `.Alt("")` appended to the chain).
- Update the apply-and-compare tests in `tests/Rask.Generators.Tests`.
- Update the two rows in `docs/diagnostics.md`'s quick-fix table.

pal-tamas:
Fixing this turned up something considerably worse than the quick-fix wording, so recording it here.

**Three analyzers were silently dead on the chain** — the syntax the framework teaches. Each identified its subject as "a static method on a class named `Generated`", which a chain never is: a chain's steps are extension methods on `Build<T>`, and its shortest spelling is a bare entry that is not an invocation at all.

| ID | What was unreported on a chain |
|---|---|
[...]

PR #705, opened 2026-08-17
fix(generators): three more analyzers were blind to the chain, found by auditing the rest

pal-tamas:
> **Stacked on #701** (→ #699 → #696) — based on `fix/rask034-chain`. Retarget down the stack as each parent merges.

#699 and #701 fixed the four analyzers that **announced themselves**: they matched *"a static method on a class named `Generated`"*, so grepping for that name found them. I said in #697 it was worth auditing the rest. It was.

## Three more, from a different family

These never mention `Generated`. They test a **TYPE**, and a chain hands back `Build<T>` rather than `T` — so nothing pointed at them. They turned up only by asking the same question of every analyzer.

| ID | Severity | What was unreported on a chain |
|---|---|---|
| **RASK032** | **Error** | Native chrome nested in HTML — `Div[NativeHeaderBar…]` serializes to nothing on a device. It is an Error precisely so that never ships, and it fired on none of the taught syntax. |
| **RASK019** | Error | Children passed to `Head[…]`, which are dropped silently. This analyzer also had **no tests at all**. |
| **RASK021** | Warning | A root rendering the page shell. It scans `Render()` for *invocations*, and a chain invokes nothing — `Html[…]` is an element access, `Doctype` a bare identifier. |

## The fixes

RASK019 and RASK032 are one-line unwraps through `BuilderEntry.ChainedComponent`, which returns `T` for a `Build<T>` and the type unchanged otherwise — safe applied unconditionally.

RASK021 keeps its **name-based** design (its existing tests declare same-named local helper methods, so a semantic test would break them) and gains the two chain spellings. The standalone-identifier arm is deliberately narrow — it excludes any identifier that merely *names* a larger expression — so a local called `Body` cannot trip it.

## ⚠️ A failing analyzer test can be a false negative

This cost me a wrong conclusion, and it is the most reusable thing in this PR.

These harnesses build a `CSharpCompilation` and **never check that it compiled**. A snippet that does not bind reports nothing — which reads exactly like a blind analyzer.

My first RASK032 test wrote `Div[NativeHeaderBar.Title("Hi")]` and failed with **`CS0103`**: a native component's *entry* is injected by the generator, which no analyzer harness runs. Rask.Core's own entries (`Div`, `Li`, `Img`, `Head`, `Html`) do bind — they are inherited from the compiled `RaskMarkup` — but anything outside Rask.Core does not.

The test now expresses the type directly with `default(Build<NativeHeaderBar>)`, identical to what a chain hands the indexer, and the new `HeadChildrenAnalyzerTests` **asserts its snippets compile** before trusting any result.

Every fix was confirmed red before green — RASK032's by stashing the fix and re-running it.

## Testing

- `dotnet format --verify-no-changes` — clean (exit 0).
- `CI=true dotnet build Rask.slnx -warnaserror -m:1` — 0 warnings, 0 errors. No new in-repo findings, which is expected: nothing nests native chrome in HTML, passes `Head` children, or renders the shell from a root.
- `scripts/run-unit-local.sh` — **6,082 passed, 0 failed**.
- Browser E2E + CLI build gate + watch hot-reload gate via the pre-push hook.

## Changelog

`[Unreleased] → Fixed`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -519,7 +519,12 @@ them until tagged releases begin.
   - **RASK027 (both the sync and async handler set) never fired on a chain.**
     `Button.OnClick(…).OnClickAsync(…)` silently dropped the async handler, which is the whole reason the
     diagnostic is an **Error** on a factory call.
-  - RASK034 has the same defect and is **not** fixed here; it is tracked separately.
+  - **RASK034 (a `BsDataGrid` column with no `Field`) never fired on a chain either.** The column
+    chooser addresses a column by the token read off `Field`, so a column without one can never be
+    shown, hidden or reordered — it sits pinned with no menu row, silently, which is exactly the
+    failure this diagnostic exists to catch. Verified against the real `Rask.Bootstrap` types rather
+    than the test's stand-ins: removing a `Field` from a sample's grid makes it fire on the offending
+    column and nothing else.
   - The chain branch is deliberately additive: the factory branch still owns `Generated.X(…)`, so nothing
     reports twice. Each new branch requires the operation's type to genuinely be `Build<T>` — without that
     the children indexer qualifies too (it is also a property reference) and every keyless row reported
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -506,6 +506,30 @@ them until tagged releases begin.
   the harness can show. The remaining arguments for splitting are size, per-file pragmas and blast
   radius — not latency; the outbox can never move, since it commits with the business change by design.
 
+### Fixed
+- **Three analyzers were silently dead on the chain — the syntax the framework teaches.** Each identified
+  its subject as "a static method on a class named `Generated`", which a chain never is: a chain's steps are
+  extension methods on `Build<T>`, and its shortest spelling is a bare entry that is not an invocation at
+  all. The build stayed green throughout, because a diagnostic that never fires breaks nothing.
+  - **RASK023 (`Img` without `Alt`) never fired on `Img.Src("/logo.png")`.** An accessibility guard
+    (WCAG 1.1.1) that was absent from every chain ever written.
+  - **RASK022 (list item missing a `Key`) never fired on a keyless chain row.** `_items.Select(i => Li[…])`
+    — the exact shape the guides teach — went unreported, so those lists reconcile by position and lose
+    focus and input state on insert/remove/reorder.
+  - **RASK027 (both the sync and async handler set) never fired on a chain.**
+    `Button.OnClick(…).OnClickAsync(…)` silently dropped the async handler, which is the whole reason the
+    diagnostic is an **Error** on a factory call.
+  - RASK034 has the same defect and is **not** fixed here; it is tracked separately.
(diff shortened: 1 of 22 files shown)


pal-tamas:
Pushed `24cc8343` — this PR now carries more than its title. Running the review step of the ship gate on the whole stack turned up **real defects in the four PRs below it**, including one that silently destroyed code. Recording them here rather than quietly folding them in.

### The destructive one

**RASK027's quick-fix deleted whatever enclosed the chain.** #699 made the analyzer fire on chains and never touched its fix, which is a `RaskCodeFixProvider<ArgumentSyntax>` — so from a chain diagnostic it walked *upward* to the nearest enclosing argument and removed it whole:

```csharp
Wrap(Content: Button.OnClick(a).OnClickAsync(b)["x"], Label: "hi")   →   Wrap(Label: "hi")
```

The component gone, and the result still compiling. The diagnostic is now anchored on the **step name** rather than the whole chain — an invocation's span starts at the *head* of the chain, which is precisely what let the walk escape — and the provider splices out that one step and never walks past the node it was handed.

### The other two fixes

- **RASK023's** could land the `Alt` on the wrong call: `Wrap(Img)` → `Wrap(Img).Alt("")`, which doesn't compile and leaves the image with no text alternative. It now acts only on the call the flagged node is the **callee** of, and the chain climb follows only links where the node is the receiver.
- **RASK014's** could trade its error for a worse one. The bare entry it writes binds only inside a **markup host** (entries are `protected static` on `RaskMarkup`), so in a service or plain class `Widget` names the type and the rewrite is `CS0119`. It is withheld there. The previous test proved nothing on this point — it asserted on text alone, *inside a plain class*.

### A false positive of my own

RASK021's bare-identifier arm (added in this PR) matched shell names on the identifier alone, so `var Body = email.Body` in a root's `Render()` raised RASK021 on ordinary code — **build-breaking under `-warnaserror`**. Narrowed to `Doctype`, the one part of the shell a chain writes bare; the other three are still caught as element accesses. A semantic check would be better and isn't available: **RS1030** forbids `Compilation.GetSemanticModel` in an analyzer, and this root's `Render()` generally lives in a different tree from the entry point being analysed.

### Perf + duplication

The four analyzers' copies of a `Build<T>` unwrapper collapse into one `BuilderEntry.BuildOf`, comparing the resolved symbol instead of calling `ToDisplayString()` on every arity-1 generic. These branches register on `OperationKind.PropertyReference` — *every property read in the compilation* — so the old shape allocated a string per `List<T>`/`Task<T>`/`Nullable<T>` read, per keystroke, in the IDE.

### Testing

Each defect has a regression test naming the shape it prevents. `dotnet format` clean; `-warnaserror` build clean; **6,087 unit / 0 failed**; **64/64 browser E2E**; CLI build gate; watch hot-reload gate.

### One thing for whoever merges

The stack is **4 commits behind `main`** (#698/#700/#702/#703 landed while this was in flight). `CHANGELOG.md` and `docs/diagnostics.md` are the likely conflict points — #700 added RASK046 and I edited that file's tables. #696 targets `main` directly and will want a rebase; the rest target each other and are unaffected until then. I've deliberately not rebased four pushed branches with open PRs.

PR #706, opened 2026-08-17
feat(native): render pure-native screens with no WebView, mixable per route

pal-tamas:
## Summary

`NativeScreen` is the pure-native counterpart of `NativeWebView` and sits in the same slot — a sibling of the native bars. Inside it, twelve new components describe a `UIView`/`android.view.View` tree instead of HTML: `NativeStack`, `NativeScroll`, `NativeList`, `NativeLabel`, `NativeButton`, `NativeTextField`, `NativeSwitch`, `NativeImage`, `NativeActivityIndicator`, `NativeDivider`, `NativeSpacer`.

**An app mixes both, per route.** One tab can be an HTML page and the next a pure-native screen. Each frame is classified by what it rendered and paints through the WebView or the native surface accordingly.

```csharp
protected override Component? Render() =>
[
    NativeHeaderBar.Title("Profile"),
    NativeScreen[
        NativeStack.Spacing(12).Padding(16)[
            NativeLabel.Text($"Signed in as {_user.Name}"),
            NativeTextField.Value(_note).OnInput(v => _note = v),
            NativeButton.Text("Save").OnClickAsync(SaveAsync)]],
    NativeTabBar.Tabs([...])
];
```

## How it works

**One render walk feeds both surfaces.** The serializer already handed every user component it walked to the session, for the native bars. That report is now a balanced **enter/exit pair**, which is enough to rebuild a whole view *tree* from the same single walk that produces the page HTML — no second render pass, and `Rask.Core` still references no `Rask.Native` type. Components that aren't native are transparent to the tree, so a screen factors into ordinary components.

**Two invariants make switching free**, and both are pinned by tests:

- A native frame **never pushes HTML**. Otherwise the WebView's DOM stops matching the HTML diff baseline (`_lastSentBuffer`) and returning to a web route repaints it from scratch.
- A backend **only hides** the content view it isn't showing, never destroys it. So a web route keeps its DOM, scroll position and JS state, and a native route patches its retained view tree instead of remounting. Both diff baselines stay truthful because the views they describe still exist.

**Routing needed no new code.** `Router`/`Outlet` render no HTML of their own, so `NativeScreen[Router]` gives `[Route]` pages, parameters, guards and type-safe `Features.Routes.*` navigation.

**Async events throughout.** Every callback has an `OnXAsync` form, awaited *before* the frame is built, so state set after an `await` paints in that frame rather than a later one.

**Opt-in and backward compatible.** Register an `INativeSurface` on `host.Services` exactly like `INativeChrome`; with none registered the family is inert and every frame paints through the WebView.

## Also fixes: RASK032 was blind to the chain syntax

It compared the argument's type against `NativeComponent`, but a chain expression is a `Build<T>` — so a native bar nested in HTML compiled clean on exactly the syntax the docs teach. It now sees through `Build<T>` on receiver and children alike and classifies by the **container**, which is also what lets native views compose legitimately inside a screen.

**RASK046** is the new mirror-image rule: HTML inside a native screen, which would otherwise silently render nothing.

## Scope — what this does NOT include

**The iOS (`UIKit`) and Android (`View`) implementations of `INativeSurface` are not written yet**, so a `NativeScreen` paints nothing on a real device today. This PR is the pipeline, the component family, the contract and the mixed-surface switching. That limitation is stated in `docs/native.md`, `NUGET.md`, `llms.txt` and the CHANGELOG rather than left implied.

## Testing

- **26 new unit tests** — 13 for the tree differ, 13 for the session. The test double retains a tree and *applies* patches the way a real backend must, so a patch list that looks right but replays wrong fails here.
- **Both load-bearing guards mutation-tested** rather than trusted on a first-try pass: disabling the emit gate turns 2 tests red; removing the native-frame commit branch turns 10 red.
- Full local gate green: `dotnet format` + warnings-as-errors build + all solution tests (`scripts/run-unit-local.sh`).
- Browser E2E green, 64/64 (`scripts/run-e2e-local.sh`). One earlier run failed a single journey intermittently and passed on a clean re-run.

## Benchmarks

`HtmlSerializerLiveRootBenchmarks`, before/after:

| Method | Allocated (before) | Allocated (after) | Mean (before) | Mean (after) |
| --- | --- | --- | --- | --- |
| RenderPageSmall | 3.98 KB | 3.98 KB | 2.066 μs | 2.043 μs |
| RenderPageLarge | 92.27 KB | 92.27 KB | 49.998 μs | 67.856 μs |
| RenderPageXLarge | 565.59 KB | 565.59 KB | 158.636 μs | 218.010 μs |

**Allocated is identical**, which is the evidence that matters for the render-walk change — it adds one predictable branch per user component and no allocation. `RenderPageSmall`, the tightest and least GC-dominated case, is unchanged within noise. The Large/XLarge means are **not** a measured regression: those runs had StdDev of 14 μs and 58 μs against 1.75 μs and 5.46 μs on the baseline, so they were too noisy on this machine to separate. Worth re-measuring on a quiet machine before drawing any conclusion from them.

## Changelog

`[Unreleased]` updated under Added and Fixed.


pal-tamas:
Superseded by #713 — same work rebased onto current main, conflict-free, with RASK046 renumbered to RASK047 (046 went to the Key-opens-the-chain rule) and the RASK032 fix switched to the shared BuilderEntry.ChainedComponent.

PR #708, opened 2026-08-17
feat(native): paint pure-native screens on iOS, behind a shared patch replay

pal-tamas:
> **Stacked on #706** — review that first. This PR's diff is against `worktree-pure-native-components`; retarget to `main` before merging the parent.

## Summary

`RaskWkWebView` now implements `INativeSurface`, so a `NativeScreen` paints as a real UIKit tree in the same container that already holds the WebView and the native bars.

The container keeps **both** content views and only toggles which is visible. That's the invariant the session depends on: while a native route is showing, the WebView's DOM still matches the HTML diff baseline (so coming back doesn't reload the page), and while a web route is showing, the retained `UIView` tree still matches the node baseline (so going back to native patches instead of re-mounting).

## The interesting part: the risky half isn't in the platform head

Patch replay is where an ordering slip shows up as a **scrambled screen** while the patch list still reads perfectly correctly — and it's exactly the code a device would normally be needed to exercise. So it isn't in the iOS backend:

- **`NativeSurfaceHost<TView>`** owns the retained tree and replays patches in order — mount, path resolution, insert/remove/move/replace.
- **`INativeViewOps<TView>`** is all a platform head supplies: create a view, set one prop, reorder children.

That puts replay on plain `net10.0` under unit test, and reduces the upcoming Android head to a mapping table.

## Layout

Auto Layout throughout — every container is a `UIStackView`, so sizing falls out of the leaves' intrinsic content sizes and there's no frame math to keep in step with the bars. `NativeScroll` and `NativeList` wrap a stack inside a scroll view, which is why the child ops consult the parent's *kind* rather than always using the view they were handed.

## Testing

- **7 new unit tests** for `NativeSurfaceHost` — mount order, prop set/unset, replace-swaps-the-view, ordered insert/remove/move replay, move-keeps-the-instance, nested path resolution, and patch-before-mount failing loudly.
- **Mutation-tested**: swapping the move op's from/to turns both move tests red, so they're load-bearing rather than incidentally green.
- Full local gate green: `dotnet format` + warnings-as-errors + all solution tests (351).
- iOS head compiles clean against the real SDK: `dotnet build src/Rask.Native/Rask.Native.csproj -p:RaskNativeHeads=ios`.
- Browser E2E run separately (this PR's code is iOS-only or native-session-only, so it can't reach the browser hosts — but verified rather than asserted).

## Honest status

**The iOS backend has not been run on a simulator or device.** It compiles against the real SDK and follows the contract the shared host is tested against, but nothing here proves it paints correctly on screen. `docs/native.md` says so plainly. Running it on a simulator is the obvious next step.

**Android is still unwritten** — that workload isn't installed on this machine, so that head would be compile-checkable only in CI.

## Correction to #706

#706 documented `NativeList` as a recycling `UITableView`/`RecyclerView`. **It isn't.** Every row is a real view, built once and kept — which suits the tens-of-rows lists most screens have, not thousands. A recycling collection's data-source model doesn't fit a patch-addressed tree, so cell reuse is a tracked follow-up rather than something the docs should have implied was already there. Fixed in the component XML docs, `docs/native.md` and the CHANGELOG.

## Changelog

`[Unreleased]` updated — the iOS backend, the shared host split, and the `NativeList` correction.


pal-tamas:
Superseded by #713, which carries this commit rebased onto current main.

Issue #685, opened 2026-08-14
A child's identity is its ordinal, so inserting a sibling re-creates every later child


pal-tamas:
A child component's identity inside its parent is its **ordinal among all entry-built children**, not
its type-plus-position and not its `Key`:

```csharp
// Component.GetOrCreateChild
var key = (typeof(T), Live.ChildPositions++);   // ChildPositions is ONE counter, shared by every type
```

`Live.ChildPositions` is reset to 0 at the top of each `RenderForLive()` and incremented once per
`GetOrCreate` call, in render order. So the slot a child lands in depends on **how many children were
built before it in the same parent's `Render()`** — across all types. Change that count and every child
after the change point misses its lookup in `PreviousChildren` and is **re-created**: new instance, new
`LiveState`, empty render cache, `OnMount` again, and any private field state gone.

`Key` does not participate. It is the diff codec's reconciliation identity (`data-rask-key`), and the
parent's child map never reads it.

### Where it was observed

Instrumenting the showcase (`RenderForLive`'s cache check, printing each node's flags per frame): typing
into the sidebar filter changes how many nav links `ShowcaseLayout` builds ahead of its `Outlet`, and the
`Outlet` comes back on the next frame with `cached=False` — a fresh instance — while `Router` and
`ShowcaseLayout` above it are `cached=True`. With the sibling count held constant the same `Outlet` is
`cached=True`, so the churn tracks the count, not the re-render.

That churn is also what turned the render-cached-`Router` bug (fixed on the builder-surface branch by
giving `Router` `BypassRenderCache`) from latent into a crash: the fresh `Outlet` had to render, in a
frame where the cached `Router` had not published `ctx.Route`. Fixing the `Router` fixed the crash; this
issue is the other half, and it is **older than the builder surface** — `main` behaves identically,
checked by running the same instrumented build there.

### Why it matters beyond that

The general shape is a keyed list with an insert at the top:

```csharp
foreach (var item in items)          // items.Insert(0, newItem) between renders
{
    yield return TodoRow.Item(item).Key(item.Id);
}
```

Every row after the inserted one shifts by a slot, so each one is handed the *next* row's instance —
which is exactly the state-follows-position bug `Key` exists to prevent, one layer below where `Key` is
consulted. Anything a row holds itself (an open/closed toggle, an edit buffer, a subscription taken in
`OnMount`) moves with the position rather than with the item.

### What a fix probably looks like

[...]

PR #710, opened 2026-08-17
refactor(html)!: move the HTML/SVG element family from Rask.Core into Rask.Html

pal-tamas:
## Summary

Moves the HTML/SVG element family out of `Rask.Core` into a new **`Rask.Html`** assembly — ~155 tag
components (`Div`, `Span`, `Table`, `Input`, the 41 `<svg>` elements, `Doctype`) in the
`Rask.Html.Components` namespace.

`Rask.Core` keeps only what its own engine constructs: `Text`, `Raw`, `Fragment`, `ErrorBoundary`,
`Context`, `Mount`, the default error/not-found pages, and the shell tags those build from (`Html`,
`Body`, `Head`, `A`, `Div`, `H1`, `P`, `Pre`, `Code`, `Details`, `Summary`, `Meta`, `Title`,
`Button`). The dependency is one-way and now checkable: Core can no longer reach for a tag component
by accident, and the DOM surface it genuinely depends on is an explicit list rather than an
accident of co-location.

Packaging is unchanged for consumers. `Rask.Html` is `IsPackable=false` and bundled into the
`Rask.Server` / `Rask.Wasm` / `Rask.Native` package `lib/` folders next to `Rask.Core.dll`, exactly
like `Rask.Client`.

## Why the namespace changed

Not cosmetic — forced. The generator emits one `public static partial class Generated` per
compilation, and `Rask.Core` still declares components. Sharing `Rask.Core.Components` would put two
public `Rask.Core.Components.Generated` types in the reference graph, which is CS0433 at every call
site. `[assembly: RaskFactoryNamespace("Rask.Html.Components")]` — the same opt-in `Rask.Native`
already uses — makes a consuming app's generator surface the factories, so `Div.Class("card")` and
`Generated.Img(…)` still resolve with no per-file `using`.

`Doctype` moves with the family (it is an HTML component). `HtmlSerializer` matches the
`DoctypeComponent` base Core keeps, so it still emits the declaration without depending on the
assembly downstream of it. The root path uses a Core-internal `CoreDoctype` rather than a bare `new`,
because that node needs a builder **entry** for stable identity — a `new` renders identical HTML and
then silently defeats the render cache on the component that re-renders every frame.

## Three generator defects this surfaced

All three were latent only because the tag entries used to arrive by **inheritance** from
`RaskMarkup` — which needs no emission in the consumer and does not care about nesting.

1. **A nested component could never be injected into**, so it silently lost the chain. The generated
   file now re-opens each enclosing type as a `partial` around the host, and RASK036 names the type to
   fix when one of them is not `partial` — instead of dropping the surface in silence. 536 nested
   components exist in this repo alone.
2. **The per-host collision filter had nothing to filter against.** `EntryHostDecls` built every host
   declaration from a `Candidate` without its member names, so an injected entry could collide with a
   member the host declares or inherits (`Style`, `Data`, `Label`, `Cite`, `ClipPath`) — CS0102/CS0108
   the moment anything is injected rather than merely shadowed.
3. **The playground never turned the builder surface on.** It only worked because the entries were
   already compiled into the referenced `Rask.Core.dll`.

`RaskBuilderEntryInjection` (MSBuild, opt-out) is new: the assembly still publishes its
`RaskEntries{Assembly}` class for referencing compilations but does not inject its own entries back
into its own hosts. A library that IS the entry set needs that split — injecting each of ~155 tags
into every other is O(n²) generated members that collide head-on with what those hosts inherit from
`Element`.

## Testing

- `dotnet build Rask.slnx` — clean.
- `CI=true dotnet build Rask.slnx -warnaserror -m:1` — **0 errors, 0 warnings**, WASM samples included.
- **`scripts/run-cli-build-e2e.sh` — 27/27.** This is the only gate that crosses the package boundary:
  it packs this commit's 15 packages to a local feed, scaffolds projects with `rask new`, restores them
  against that feed and builds with `-warnaserror`. It is what proves the `Rask.Html` bundling into the
  three host `lib/` folders actually works for a *restored consumer*, and that `RaskBuilderEntryInjection`
  is right at its default there — an in-repo build uses project references and could not catch either.
- `dotnet format --verify-no-changes` — clean.
- Full unit suite — green. Notables: `Rask.Core.Tests` 2018, `Rask.Html.Tests` 461 (new),
  `Rask.Generators.Tests` 425, `Rask.Cli.Tests` 796, `Rask.Bootstrap.Tests`, `Rask.Server.Tests`.

All of the above was re-run after merging current `main` (through #714), which brought #698's MDN docs,
#700's keyed-child identity, #702, #703's breaking forms change and #705/#707's chain-aware analyzers.

### Benchmarks

`*RenderRoundTrip*`, Release, `[MemoryDiagnoser]`, baseline taken from this branch's merge base
(`ef63304f`) in a separate checkout. `Allocated` is byte-for-byte unchanged across all five —
what a source-level move should look like on the render hot path:

| Benchmark | `ef63304f` | this branch | delta |
|---|---:|---:|---:|
| `RenderOnce` | 81.2 KB | 81.2 KB | 0 |
| `RenderTenTimes` | 151.51 KB | 151.51 KB | 0 |
| `RenderKeyedList100_ShuffledEachIteration` | 94.55 KB | 94.55 KB | 0 |
| `RenderDeep_50UserComponents` | 76.68 KB | 76.68 KB | 0 |
| `RenderAndBuildPayload` | 35.31 KB | 35.31 KB | 0 |

`Mean` is not reported: StdDev exceeded the between-run differences on both sides (the machine was
running other builds), so only `Allocated` carries signal here.

### Browser E2E

`scripts/run-e2e-local.sh` — **64/64 passed** across 21 shards, including every WASM one:
`WasmExampleTests` (10), `CrdtExampleTests` (4), `BrowserJobsWasmExampleTests` (3),
`PlaygroundExampleTests`, `SiteExampleTests`, `StandaloneWasmExampleTests`,
`WasmSubPathExampleTests`, `WasmWatchHotReloadTests` and both WASM auth suites — alongside the
server side (`ServerExampleTests`, `ShopExampleTests`, the EfCore/Sqlite journeys and
`MorphSelectSelectionTests`).

So the moved element family is exercised through both runtimes: server-rendered over WebSocket and
browser-WASM over JSImport/JSExport.

## Test-suite changes worth a reviewer's eye

Two places where the move would otherwise have quietly *reduced* coverage rather than broken anything:

- **`ComponentDocumentationTests` walked only `src/Rask.Core/Components`.** After the move that would
  have silently stopped checking ~150 tags — staying green forever while documenting nothing. It spans
  both projects now, which is why `Rask.Core.Tests` went 1881 → 2019.
- **The analyzer harnesses never ran the generator.** They did not need to: the framework's builder
  entries were precompiled into `Rask.Core.dll` and arrived by inheritance. A referenced library's are
  injected, so after the move a chain over a moved tag failed to bind and the analyzer saw nothing to
  report — four of #705/#707's chain tests went red for that reason alone. The harnesses run the
  generator now, matching a real build.

## Sibling test project

`tests/Rask.Html.Tests` (461 tests) — 112 component suites moved with the components they cover; the 32
left in `Rask.Core.Tests/Components` cover what `Rask.Core` still declares. 2317 total before, 2479 after
(the difference is the widened documentation theory), so the split moved tests rather than losing them.

## Breaking changes

- Tag components move from `Rask.Core.Components` to `Rask.Html.Components`. Code that names a tag
  **type** explicitly needs `using Rask.Html.Components;`. The chain and the factories are unaffected.
- A type that **encloses** a component must be declared `partial`; RASK036 reports it by name. This is
  the price of a nested component being injected into rather than inheriting the tag entries.


pal-tamas:
Attempted this merge against current `main` while landing #714 (Page/Screen). Got ~95% of the way and stopped at one file. **Nothing was pushed — this branch is untouched at `4c4887ee`.** Posting the findings because two of them are silent regressions that survive a green build.

## Two regressions in the branch as it stands

**1. It reverts #698's MDN docs for the whole HTML element family.**

The moved copies carry no doc comments, while `main` documented the originals:

```
origin/main   src/Rask.Core/Components/Abbr.cs   -> full MDN <summary>
this branch   src/Rask.Html/Components/Abbr.cs   -> no docs at all
```

`git merge origin/main` produces 63 modify/delete conflicts (deleted here, documented on main). Resolving them the obvious way — accept the deletion — drops documentation for ~155 components with nothing to flag it.

**2. It deletes the RASK046 analyzer.**

`src/Rask.Generators/Analyzers/KeyOpensChainAnalyzer.cs` (140 lines) and `tests/Rask.Generators.Tests/KeyOpensChainAnalyzerTests.cs` are absent from the branch, and `RASK046` appears nowhere in it — while `docs/diagnostics.md` still documents RASK046. The file exists at the merge base (`8c36ada8`), so this looks like a bad resolution rather than intent.

## The reconciliation, if it's useful

Available locally as branch `move-html-merge` (`1d2cb9c8`) with a patch at `.claude/rask-html-merge-wip.patch` — 485 files, +12969/-859.

- All 150 moved components regenerated from main's **documented** versions, re-applying the two structural changes the move makes (namespace prefix, `partial`). Audited first: across all 151 moved files the move changes only namespace / `partial` / dropped `new`. `Doctype.cs` is the one exception (its added note about the Core/Html split) and was left as authored.
- RASK046 analyzer + tests restored from main.
- Resolved: the 20 `Rask.Bootstrap` files, both Playground files, `BuilderEntry.cs`, `docs/diagnostics.md`, `docs/elements.md`, `CLAUDE.md` (union of the new `Rask.Html` bullet and main's updated generator line; `add-html-tag` now points at `tests/Rask.Html.Tests/`), `CHANGELOG.md` (this branch's `### Changed` spliced between main's `Added` and `Fixed`).
- `Element.cs` by hand: main's doc comments, this branch's `new`-free `Style`/`Data`, and the explanatory comment about why `new` is gone, kept verbatim.

## Where it stops

`src/Rask.Generators/ComponentFactoryGenerator.cs`. This branch changes it +149/-94 across 22 scattered hunks; `main` changes it +482/-79 in the same regions (main's `Key`/`ClaimKey` work from #685/#700 lands right where these hunks are). Git auto-merges most of it into something incoherent — main's call sites against this branch's `Candidate` record — which the build catches:

```
24x CS1061  Candidate does not contain 'EnclosingAllPartial' / 'EnclosingTypes'
16x CS0103  the name 'seedCarriesKey' does not exist
```

Reconciling that means holding both redesigns of the code that emits every component's chain surface. That belongs with whoever wrote the move, not with a merge heuristic.

## Also

`main` has moved twice since: **#720** (fails the build when two analyzer assemblies share a diagnostic id — note it only catches *cross-assembly* collisions, since Roslyn's `RS1019` is per-compilation) and **#719** (native screens). `main` is at `b2fca9c6`, so this needs re-merging regardless.

PR #711, opened 2026-08-17
feat(native): paint pure-native screens on Android too

pal-tamas:
> **Stacked on #708** (which is stacked on #706) — review those first. Retarget to `main` before merging the parents, or this never gets CI (`ci.yml` only triggers on PRs targeting main).

## Summary

`RaskAndroidWebView` now implements `INativeSurface` — the mirror of the iOS backend from #708. A `NativeScreen` paints as a real `View` tree in the same container that already holds the WebView and the native bars, and the container keeps **both** content views, toggling only which is `Visible`.

That's the same invariant as iOS, and it's what keeps the session's two diff baselines truthful: the WebView's DOM still matches the HTML baseline while a native route shows, and the retained `View` tree still matches the node baseline while a web route does.

## Why this head is small

`NativeSurfaceHost<TView>` (#708) already owns the retained tree and the ordered patch replay, unit-tested on plain `net10.0`. So this is only a mapping table from `NativeNodeKind`/`NativePropId` onto widgets — which was the whole point of that split.

**Framework widgets only** — no AndroidX or Material dependency, matching how the chrome bars are already built. A pure-native screen themes with the app's default theme and adds nothing to the APK.

## Two things Android needed that iOS didn't

- **`LinearLayout` has no per-child spacing and no cross-axis stretch.** `RaskStack` owns both and re-applies them to its children whenever either the setting or the child list changes — so spacing survives an insert, a remove and a move.
- **Controlled inputs need a suppress flag.** Writing `Text`/`Checked` back from a frame fires the change event; without the guard it would echo straight back as if the user had typed it, and loop.

## Testing

- Compiles clean against the real Android SDK + JDK, both TFMs (`net10.0` and `net10.0-android`).
- **Verified the compile actually covers the new file** rather than trusting a green build: `Platforms/**` is `Compile Remove`d per-TFM, so I inserted a deliberate type error into `AndroidViewOps.cs`, confirmed it failed the `net10.0-android` compile, and removed it.
- Full local gate green: `dotnet format` + warnings-as-errors + all solution tests.

Local build needs explicit SDK/JDK paths on this machine (the SDK isn't where dotnet looks by default, and the JDK path must be the inner `Contents/Home`):

```bash
dotnet build src/Rask.Native/Rask.Native.csproj -p:RaskNativeHeads=android \
  -p:AndroidSdkDirectory=/opt/homebrew/share/android-commandlinetools \
  -p:JavaSdkDirectory=/opt/homebrew/opt/openjdk@17/libexec/openjdk.jdk/Contents/Home
```

## Honest status

**Neither backend has been run on an emulator, simulator or device.** Both compile against their real platform SDKs and both follow the contract the shared host is tested against, but nothing here proves either paints correctly on screen. `docs/native.md` and the CHANGELOG say so plainly. Running the showcase on an emulator is the obvious next step, and the on-device Appium suite (`tests/Rask.Native.Appium.Tests`) is where a real assertion would live.

## Changelog

`[Unreleased]` updated — both backends now ship, with the not-run-on-device caveat.


pal-tamas:
Superseded by #713, which carries this commit rebased onto current main.

PR #713, opened 2026-08-17
feat(native): pure-native screens with no WebView, mixable per route, on iOS and Android

pal-tamas:
Replaces #706, #708 and #711 — same work, rebased onto current `main`, conflict-free, with the RASK032 duplication resolved. Three commits, one per layer, so it still reads as a stack.

## What this adds

`NativeScreen` is the pure-native counterpart of `NativeWebView` and sits in the same slot. Inside it, twelve new components describe a `UIView` / `android.view.View` tree instead of HTML: `NativeStack`, `NativeScroll`, `NativeList`, `NativeLabel`, `NativeButton`, `NativeTextField`, `NativeSwitch`, `NativeImage`, `NativeActivityIndicator`, `NativeDivider`, `NativeSpacer`.

```csharp
protected override Component? Render() =>
[
    NativeHeaderBar.Title("Profile"),
    NativeScreen[
        NativeStack.Spacing(12).Padding(16)[
            NativeLabel.Text($"Signed in as {_user.Name}"),
            NativeTextField.Value(_note).OnInput(v => _note = v),
            NativeButton.Text("Save").OnClickAsync(SaveAsync)]],
    NativeTabBar.Tabs([...])
];
```

**An app mixes both, per route** — one tab an HTML page, the next a pure-native screen. Each frame is classified by what it rendered and paints through the WebView or the native surface.

## The three commits

**1. Pipeline + components.** One render walk feeds both surfaces: the serializer already reported each user component it walked (for the native bars), and that report is now a balanced **enter/exit pair** — enough to rebuild a whole view *tree* from the same walk that produces the page HTML. No second render pass, and `Rask.Core` still references no `Rask.Native` type. Routing needed no new code: `Router`/`Outlet` render no HTML of their own, so `NativeScreen[Router]` gives `[Route]` pages, params, guards and type-safe `Features.Routes.*`. Every callback has an awaited `OnXAsync` form, awaited *before* the frame is built.

**2. iOS backend + shared patch replay.** The half that could hide a bug is deliberately not in the platform head: `NativeSurfaceHost<TView>` owns the retained tree and replays patches in order, and a head supplies only an `INativeViewOps<TView>` mapping table. Patch replay is where an ordering slip shows as a scrambled screen while the patch list still reads correctly — so it runs on plain `net10.0` under unit test rather than needing a device.

**3. Android backend.** The same contract, framework widgets only (no AndroidX/Material), matching how the chrome bars are built.

## Two invariants worth reviewing closely

- **A native frame never pushes HTML.** Otherwise the WebView's DOM stops matching the HTML diff baseline and returning to a web route repaints it from scratch.
- **Neither content view is ever destroyed, only hidden.** That's what keeps both diff baselines truthful — the WebView keeps its DOM, scroll position and JS state, and a native route patches its retained tree instead of remounting.

## RASK047, and RASK032

**RASK047** (new): HTML nested inside a `NativeScreen`, which would otherwise silently render nothing. It's 047 rather than 046 — 046 went to your Key-opens-the-chain rule while this was in flight.

**RASK032** was blind to the chain (it compared against `NativeComponent`, but a chain hands back `Build<T>`). Fixed here via the **shared `BuilderEntry.ChainedComponent`** rather than a private copy, and it now classifies by the *container*, which is what lets native views compose legitimately inside a screen. If #705 lands first this becomes a trivial conflict — the mechanism is the same one.

## Testing

- **33 new unit tests** — 13 tree differ, 13 session, 7 surface host. The test double retains a tree and *applies* patches the way a real backend must, so a patch list that looks right but replays wrong fails.
- **Every load-bearing guard mutation-tested** rather than trusted on a green first run: disabling the emit gate turns 2 red, removing the native-frame commit branch turns 10 red, swapping the move op's from/to turns 2 red.
- **The Android compile was verified to actually cover the new file** — `Platforms/**` is `Compile Remove`d per-TFM, so a green build proves nothing on its own. I inserted a deliberate type error into `AndroidViewOps.cs`, confirmed `net10.0-android` failed, and reverted it.
- Full local gate green (`dotnet format` + warnings-as-errors + all solution tests), browser E2E green, and both heads compile against their real SDKs.

**Your new RASK046 caught a real bug in my own test code** during the rebase: `NativeLabel.Text(i).Key(i)` in the keyed-list test wrote `Text` to an instance the `Key` then discarded. Fixed to `.Key(i).Text(i)`.

## Benchmarks

`HtmlSerializerLiveRootBenchmarks`, before/after: **Allocated identical** — 3.98 KB / 92.27 KB / 565.59 KB. `RenderPageSmall` 2.066 → 2.043 μs. The Large/XLarge means are **not** a measured regression: those runs had StdDev of 14 μs and 58 μs against 1.75 μs and 5.46 μs on the baseline, so they were too noisy on this machine to separate. Worth re-measuring on a quiet machine before reading anything into them.

## Honest status

**Neither backend has been run on an emulator, simulator or device.** Both compile against their real platform SDKs and follow the contract the shared host is tested against, but nothing here proves either paints correctly on screen. `docs/native.md` and the CHANGELOG say so plainly. Running the showcase on an emulator, and asserting it in `tests/Rask.Native.Appium.Tests`, is the obvious next step.

`NativeList` also does **not** recycle rows — every row is a real view, built once and kept. That suits tens of rows, not thousands; a recycling collection's data-source model doesn't fit a patch-addressed tree.


pal-tamas:
Superseded by #715. Do not merge this one: its commit carries `NativeLabel.Text(i).Key(i)` in the keyed-list test — a RASK046 violation whose fix was never staged. No CI job builds the test projects (tests run locally here), so it went green 10/10 while carrying a file that fails the pre-commit gate. Fixed in #715.

Issue #694, opened 2026-08-14
MDN gaps: <var> is missing, and <button>/<video>/<form> lack attributes the spec defines


pal-tamas:
Documenting the element surface against MDN turned up one missing element and a set of per-element attribute gaps. Element coverage is otherwise excellent — 147 tags, and the only non-experimental omission is `<var>`.

## Missing element: `<var>`

[`<var>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/var) marks a variable in a mathematical expression or programming context. Rask already has its whole neighbourhood — `Code`, `Samp`, `Kbd`, `Dfn` — so this looks like an oversight rather than a decision. Cheap to add via the `add-html-tag` skill.

Deliberately **not** proposed: `<math>` (the whole MathML namespace is a separate undertaking), `<fencedframe>` and `<selectedcontent>` (experimental, Chromium-only).

## `<button>` is missing the form-override and popover attributes

`Button` exposes `Type`, `Disabled`, `Name`, `Value`. MDN lists these as also non-deprecated:

- `form`, `formaction`, `formenctype`, `formmethod`, `formnovalidate`, `formtarget`
- `popovertarget`, `popovertargetaction`
- `autofocus`
- (newer: `command`, `commandfor`, `interestfor`)

The form-override set is the notable one, because **`Input` already has all six** (`Form`, `FormAction`, `FormEnctype`, `FormMethod`, `FormNovalidate`, `FormTarget`). So a submit button can override the form's action when written as `<input type="submit">` but not as `<button>` — an inconsistency rather than a considered omission. `popovertarget` is the other half of the `popover` global attribute in #693.

## `<video>` is missing four attributes

`Video` has `Poster`, `Width`, `Height`, `PlaysInline`, plus the shared `HtmlMediaElement` set. [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/video) also lists:

- `controlslist` — `nodownload` / `nofullscreen` / `noremoteplayback`
- `disablepictureinpicture`
- `disableremoteplayback`
- `loading` — `lazy` for a video below the fold

## `<form>` is missing `rel`

`Form<TModel>` has `Enctype`, `Target`, `AcceptCharset`, `Autocomplete`, `Novalidate`, `Name`. MDN also lists `rel`.

[...]

PR #723, opened 2026-08-17
fix(skills): make the add-html-tag test template compile, and add the <var> element

pal-tamas:
## Summary

**#722 landed the skill's path fix while this was open — but it corrected `SKILL.md` only, and left
`templates/ComponentTests.cs` untouched.** That template is what actually gets copied, and it is stale
in four ways at once:

```
// Template — copy to tests/Rask.Core.Tests/Components/{Tag}Tests.cs.   <- moved project
namespace Rask.Core.Tests.Components;                                   <- moved namespace
public class {Tag}Tests                                                 <- not partial, no RaskMarkup
    HtmlSerializer.Serialize(new {Tag}())                               <- pre-chain style
```

So a reader following the corrected `SKILL.md` to the right *directory* still copies a file that does
not compile there. This rewrites it to match what the moved tests actually look like, and states the
part that is easy to get wrong and hard to diagnose:

```csharp
public partial class {Tag}Tests : global::Rask.Core.RaskMarkup
```

`partial` and the `RaskMarkup` base are load-bearing. Rask.Html's builder entries are *injected* into a
consumer's markup hosts rather than inherited from Rask.Core, so a non-partial host gets none of them
(RASK036) and `{Tag}` simply would not resolve — the failure reads as "the tag does not exist", which
points nowhere near the missing modifier.

**Also: `<var>`** — the one element MDN lists that Rask had no component for, part of #694.

## Why the two are together

The second is how the first was found. I set out only to fix the skill, and the honest way to verify a
scaffolding template is to scaffold with it rather than read it — which is what surfaced that the
template was stale in ways `SKILL.md` alone could not fix, and produced a working component on the way.

## Testing

- `CI=true dotnet build Rask.slnx -warnaserror -m:1` — 0 errors, 0 warnings.
- `dotnet format --verify-no-changes` — clean.
- Full unit suite — 49 assemblies, green (`Rask.Html.Tests` 464, `Rask.Example.Shared.Tests` 338).
- `DemoMarkupGoldenTests` caught the demo change, as designed. Regenerated with `RASK_UPDATE_GOLDEN=1`;
  the diff is one line, `+var`, which is the whole review.

## User-facing surface

- `docs/elements.md` lists `var` in **Text & inline** with its MDN link.
- `ElementsTextDemo` uses it: that demo previously wrote *"a variable x"* as plain prose, which is
  precisely the case `<var>` exists for.
- CHANGELOG entry under `### Added`.

Cleanly separable if you would rather `<var>` landed on its own — the template fix stands alone.

### DIFF
--- a/.claude/skills/add-html-tag/SKILL.md
+++ b/.claude/skills/add-html-tag/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: add-html-tag
-description: Scaffold a new HTML tag component in Rask.Html. Use whenever adding support for an HTML element (e.g. <dialog>, <details>, <progress>, <video>) to the Rask framework. Creates src/Rask.Html/Components/{Tag}.cs and the matching tests/Rask.Core.Tests/Components/{Tag}Tests.cs asserting exact attribute order; the factory is generated automatically.
+description: Scaffold a new HTML tag component in Rask.Html. Use whenever adding support for an HTML element (e.g. <dialog>, <details>, <progress>, <video>) to the Rask framework. Creates src/Rask.Html/Components/{Tag}.cs and the matching tests/Rask.Html.Tests/Components/{Tag}Tests.cs asserting exact attribute order; the factory is generated automatically.
 ---
 
 # add-html-tag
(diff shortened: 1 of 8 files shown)


pal-tamas:
Heads-up: this is already on main. I landed the same fix as #722 (`6c20dea2`) a few minutes ago, which is why this now reports DIRTY — the change it makes is already there.

#722 covered the same test-path correction plus two things this one may not, so worth checking before you close or rebase:

1. The skill's **"attributes" reference** was stale in the other direction too — it named `src/Rask.Html/Components/Button.cs`, which does not exist, because `Button` is one of the handful of tags `Rask.Core` retained for its shell and error pages. #722 repointed it to `Img.cs` and added a note about the split, so a retained tag isn't taken as the pattern to copy for a new one.
2. The same test path in **`CLAUDE.md`**, which had the same staleness.

If your branch has anything #722 missed, please rebase onto main and keep just that delta — I'd rather not have your work dropped on the assumption it's a pure duplicate. Otherwise this can be closed.

Sorry for the collision. Several sessions were merging the same queue today and I didn't announce this one before starting, which is on me — the skill gap was mentioned to me by #710's author as a follow-up and I took it straight to a branch.

Issue #721, opened 2026-08-17
CLI build gate is flaky: 1 of 27 fails intermittently and aborts the push


pal-tamas:
One of the 27 tests in `scripts/run-cli-build-e2e.sh` fails intermittently. Same working tree, same commit, two consecutive runs:

```
run 1   Total tests: 27   Passed: 26   Failed: 1     4.47 min
        pre-push: CLI build gate FAILED — push aborted. The code the CLI writes doesn't compile.
run 2   Total tests: 27   Passed: 27                 3.20 min
        ==> CLI build gate passed.
```

The tree under test was `native-screens-4` (#719) merged with `main` at `96c66eb5`. To rule out a real regression from that combination, the same gate was run on **`main` alone**, which passed 27/27. So: main passes, the merge passes on the second attempt, and the merge failed once. That is intermittency, not a broken merge.

### Why this is worth fixing rather than re-running

This gate is a **pre-push hook**. It is the only thing standing between the code `rask new` writes and the remote, and it aborted a legitimate push. The failure mode it trains is the dangerous one: a red gate that goes green on retry teaches everyone to retry rather than read, which is exactly how a genuine "the scaffolded code doesn't compile" failure gets waved through. The same argument was made for the flakes fixed in #691/#692.

### What I could not capture

The failing test's **name is not in this report**, and that is the first thing anyone fixing this will want. The first run was piped to `tail`, so only the summary survived; by the time it was re-run the failure had cleared. Whoever picks this up should run the gate in a loop and keep the **full** output:

```bash
for i in $(seq 1 10); do
  bash scripts/run-cli-build-e2e.sh > "/tmp/cli-$i.log" 2>&1
  echo "$i: $?"
done
grep -l "Failed:  *[1-9]" /tmp/cli-*.log
```

Note that `bash scripts/run-cli-build-e2e.sh | tail` reports **tail's** exit status, not the script's, so a loop that checks `$?` through a pipe will record success for every run. Redirect to a file instead, as above, and key on the script's own `==> CLI build gate passed.` line.

### Likely shape

The gate compiles freshly scaffolded projects, so the usual suspects are contention rather than logic: parallel `dotnet build` on the same NuGet cache or obj/ directory, a temp directory reused between cases, or a restore racing a build. Worth checking whether the 27 cases share a scaffold root, and whether anything in the script runs them concurrently.

pal-tamas:
[...]

PR #724, opened 2026-08-18
fix(gates): classify a build failure by error kind before blaming the branch

### LINKED ISSUE #718: CLI build gate blames the scaffolded code when the wasm-tools workload is unresolvable
## What happens

When the `wasm-tools` workload is transiently unresolvable, every browser-targeting build fails at *evaluation* with `NETSDK1147`. That takes down two gates, and the CLI one reports it as:

```
pre-push: CLI build gate FAILED — push aborted. The code the CLI writes doesn't compile.
```

That message names the wrong culprit. The scaffolded code compiles fine — the machine cannot build browser targets at that moment. Today it cost two sessions about an hour between them, chasing a scaffolder bug that did not exist.

Measured on the failing run: **0 `error CS`, 24 `error NETSDK1147`**. The same CLI gate run standalone minutes later passed **823 / 0 failed**, with no code change.

## Why the workload goes missing

It is usually not a broken install. A **concurrent** `sudo dotnet workload install <anything>` (in our case `android`, from another session on the same machine) installs a new workload *set* and bumps the shared `microsoft.net.workload.mono.toolchain.current` / `emscripten.current` manifests for the whole SDK band. Between the manifest bump and the pack restore, the manifests reference Emscripten packs that are not on disk yet, so `wasm-tools` resolves as missing — **machine-wide, in every worktree** — while `dotnet workload list` still cheerfully lists it as installed.

It clears on its own when the installing process exits. Racing it with your own `dotnet workload restore` does not help.

A stale SDK feature-band registration (workload registered for an SDK that is no longer installed, with no `global.json` pinning) produces the same `NETSDK1147` and is worth mentioning in the same breath.

## Suggested fix

In `scripts/run-e2e-local.sh` (CLI build gate) and the pre-push hook, when the build fails, count the error kinds before assigning blame:

- if `error NETSDK` > 0 and `error CS` == 0 → **do not** claim the generated code is broken. Fail with something like:

  > Your machine cannot build browser targets right now: the `wasm-tools` workload is unresolvable (`NETSDK1147`). This is not your branch.
  > Check for a workload install in flight: `ps aux | grep "workload install"` and `ls -la /usr/local/share/dotnet/metadata/workloads/InstalledWorkloadSets/` — a workload set dated in the last few minutes is the cause. Wait for that process to exit and re-run; do not run `dotnet workload restore` alongside it.

- otherwise keep the current message, which is correct when there really are `CS` errors.

The same check is worth adding wherever the E2E gate builds the sample graph, since it fails first and for the same reason.

## Why it matters

The gate is load-bearing — it is the only thing that crosses the package boundary and proves `rask new` output compiles. A gate that goes red for an environmental reason *and blames your code* trains people to distrust it, which is worse than it simply failing. Distinguishing "your branch is broken" from "your machine is busy" is a few lines of `grep -c`.

## Repro

1. In one shell: `sudo dotnet workload install android`
2. While that runs, in another: `git push` on any branch (or `dotnet build` any `net10.0-browser` project)
3. Observe `NETSDK1147`, the CLI gate blaming the scaffolded code, and `dotnet workload list` still listing `wasm-tools`

Found while landing #703; independently hit and diagnosed by a second session working the same merge queue.

pal-tamas:
## Summary

Both the CLI build gate and the E2E gate build browser targets, so both fail for reasons that have nothing to do with your branch. The commonest is `NETSDK1147`: a concurrent `dotnet workload install` anywhere on the machine installs a new workload set and bumps the shared `mono.toolchain.current` / `emscripten.current` manifests for the whole SDK band. Between the manifest bump and the pack restore they reference Emscripten packs that are not on disk yet, so `wasm-tools` resolves as missing **machine-wide, in every worktree** — while `dotnet workload list` still lists it as installed.

The CLI gate reported that as *"the code the CLI writes doesn't compile"*. Measured on the failing run: **0 `error CS`, 24 `error NETSDK1147`**; the same gate passed minutes later with no change. It cost two sessions about an hour between them, chasing a scaffolder bug that did not exist.

A gate that goes red for an environmental reason *and blames your code* trains people to distrust it, which is worse than it simply failing.

## What changed

`scripts/lib/build-failure.sh` classifies a captured build log into one of four kinds:

| kind | condition | what the gate says |
|---|---|---|
| `code` | `error CS` present | the gate's existing message — the branch is broken |
| `workload` | `NETSDK1147`, no `CS` | this is not your branch; how to confirm the install in flight |
| `sdk` | another `NETSDK`, no `CS` | an SDK/restore problem, still not the branch |
| `unknown` | neither | not a compile failure at all — a failing assertion, a timeout, a host that exited early |

`CS` wins when both appear: a workload problem does not excuse real compile errors.

Applied to the CLI build gate, the E2E gate's graph build (it fails first, and for the same reason), and both `.githooks/pre-push` messages, which now stay neutral about a culprit the script has already named.

## Testing

- `scripts/tests/build-failure-kind.test.sh` — a table test over all nine rows, run by `run-unit-local.sh` before anything else. The decision is four predicates over two counts and it decides whether a red gate sends you to your own diff or to `ps aux`, so it is stated rather than left to the two cases someone happened to try. Same reasoning as `BakeScopedAssetsTask.IsNodeReuseBakeFailure`'s table test in #690.
- **Verified by making it fail**: reordering the classifier to check `workload` before `cs` turns the "CS wins when both appear" row red; restoring it goes green. The first attempt at that break silently did not apply, and the suite stayed green — which is exactly the false negative the check exists to rule out, so it is worth stating that the red was observed.
- `scripts/` and `.githooks/` added to the pre-commit path filter. A change to the gate logic was otherwise the one change that skipped the gate.
- Full local format + unit gate green; E2E gate green on push (this PR edits `run-e2e-local.sh`, so that run is the real verification of the edit).

## Notes

`grep -c … || true` throughout, never `grep -q`: with `pipefail` a `-q` grep exits on the first match, the writer takes `SIGPIPE`, and the pipeline reports failure — the same trap already documented in `.githooks/pre-commit`, where it once let a 421-file commit through unchecked.

Closes #718

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -32,7 +32,11 @@ fi
 #
 # Leaving these out meant a whole feature could land ungated — the playground tutorial was largely a
 # samples/ + docs/ change, and reported "no code changes staged".
-if ! git diff --cached --name-only | grep -E '^(src/|samples/|docs/|tests/|benchmarks/|Rask\.slnx$|Directory\.)' >/dev/null; then
+#   scripts/ and .githooks/ are in here because the gates are themselves code with tests now:
+#   scripts/run-unit-local.sh runs scripts/tests/*.test.sh, which covers the failure classification
+#   that decides whether a red gate blames your branch or your machine (#718). Without these prefixes a
+#   change to the gate logic is the one change that skips the gate.
+if ! git diff --cached --name-only | grep -E '^(src/|samples/|docs/|tests/|benchmarks/|scripts/|\.githooks/|Rask\.slnx$|Directory\.)' >/dev/null; then
   echo "pre-commit: no code changes staged — skipping the format + unit gate."
   exit 0
 fi
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -20,8 +20,10 @@ else
   echo "pre-push: running the local E2E gate (browser journeys)."
   echo "          bypass with 'git push --no-verify' or RASK_SKIP_E2E=1 (e.g. docs-only pushes)."
 
+  # As with the CLI gate below: run-e2e-local.sh classifies a build failure by error kind and has
+  # already named the culprit, so this line stays neutral about which one it was (#718).
   if ! "$root/scripts/run-e2e-local.sh"; then
-    echo "pre-push: E2E gate FAILED — push aborted. Fix the failing journey, or bypass with --no-verify." >&2
+    echo "pre-push: E2E gate FAILED — push aborted. Fix the cause named above, or bypass with --no-verify." >&2
     exit 1
   fi
 fi
@@ -43,8 +45,10 @@ else
   echo "pre-push: this push touches the generators or the packages they emit — running the CLI build gate."
   echo "          bypass with 'git push --no-verify' or RASK_SKIP_CLI_BUILD_E2E=1."
 
+  # The gate classifies its own failure by error kind and has already said which culprit it found —
+  # a NETSDK-only failure is the machine, not the branch (#718). Don't second-guess it here.
   if ! "$root/scripts/run-cli-build-e2e.sh"; then
-    echo "pre-push: CLI build gate FAILED — push aborted. The code the CLI writes doesn't compile." >&2
+    echo "pre-push: CLI build gate FAILED — push aborted." >&2
     exit 1
   fi
 fi
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,22 @@ them until tagged releases begin.
   mathematical or programming context; not emphasis (`em`) and not literal code (`code`).
 
 ### Fixed
+- **A gate that failed because the machine could not build browser targets blamed your code for it**
+  (#718). A concurrent `dotnet workload install` anywhere on the machine bumps the shared workload
+  manifests for the whole SDK band, and for a moment they reference Emscripten packs that are not on disk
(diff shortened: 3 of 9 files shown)


pal-tamas:
Superseded by #725, which landed on main while this was in its E2E gate and fixes #718 more broadly — all four pre-push arms go through one `run_gate` helper rather than the two arms this PR wrapped, and it also addresses #721's shared-cache hazard. This branch is now `CONFLICTING` against main and its core is duplicate work, so closing rather than rebasing.

Two pieces here are **not** in #725 and are worth keeping. Opening a small follow-up with just these:

1. **`scripts/` and `.githooks/` are still missing from the pre-commit path filter.** The filter matches `src/|samples/|docs/|tests/|benchmarks/|Rask.slnx|Directory.`, so a change to the gate scripts themselves is the one change that skips the gate. #725 is a `scripts/` + `.githooks/`-only commit, so it took that path in — which is the demonstration, not a hypothetical.

2. **The verdict logic has no committed regression test.** #725 verified its classification by making it fire across four fake gates, which is the right check to have run, but it was run once by hand and the logic now lives inline in `pre-push`. Lifting it into a sourceable `scripts/lib/build-failure.sh` gets it a table test that `run-unit-local.sh` runs every time, and makes the classification apply when `run-cli-build-e2e.sh` / `run-e2e-local.sh` are invoked directly rather than only through the hook.

No objection to #725's approach — it is the better shape for the fix itself.

PR #759, opened 2026-08-18
feat(sqlite): verify the Litestream replica is restorable, not just that replication runs

### LINKED ISSUE #751: Rask.SQLite.Litestream reports that replication is running, never that the backup is restorable
## The gap

`Rask.SQLite.Litestream` can tell you that replication is *running*. It cannot tell you that what has
been replicated is *restorable*.

`LitestreamStatus` exposes `IsReplicating`, `LastStartedAt` / `LastExitedAt`, `RestartCount`,
`LastExitCode` and `LastError`, and `docs/sqlite.md` wires it to a `/health/backup` endpoint. Every one
of those is a fact about the local child process. A replica that is silently writing to the wrong prefix,
a bucket whose credentials were rotated to read-only, a `-config` file pointed at a database nobody
writes to any more — all of these keep `IsReplicating` true and `RestartCount` flat, and all of them are
discovered at the only moment that matters, which is the restore.

This is the one place the Rails story is still ahead of ours. `litestream-ruby` ships
`Litestream::VerificationJob`, which iterates the configured databases and calls
`Litestream.verify!(path)` on each — the README's suggested schedule is every 6 hours, so that you get
alerted when backups stop working rather than when you need them. We have no equivalent.

## Proposal

A verification pass that proves the round trip, off by default, on the existing seams:

1. Write a sentinel into the live database (a `__rask_backup_probe` row carrying a UTC timestamp / GUID).
2. Wait, bounded, for replication to carry it.
3. `litestream restore` to a **temporary path** and open the restored file.
4. Assert the sentinel is present, then delete the temp file.
5. Record the outcome, and make it readable next to the existing liveness.

Shape, matching what the package already does:

- `ISqliteBackupVerifier` / `LitestreamVerifier` with `VerifyAsync(ct)`, driven through the existing
  `ILitestreamExecutor` so it is testable with the fake and no binary.
- An optional hosted service on a `PeriodicTimer`, exactly like `Rask.SQLite.Snapshots` — `Interval`
  (default daily), `Enabled` default **false**.
- Extra fields on `LitestreamStatus` (`LastVerifiedAt`, `LastVerifyOutcome`, `LastVerifyError`) so the
  `/health/backup` endpoint reports "restorable at T" rather than only "running".

## Traps worth naming up front

- **`LitestreamCommand.Restore` hardcodes the output to `DatabasePath`.** Verification must restore
  *somewhere else* — it needs an output-path parameter, and in `-config` mode the current builder does not
  emit `-o` at all. That is the first thing to change, and the pure arg builder makes it cheap to test.
- **A restore costs egress.** On S3/GCS/ABS this is a real bill and a real transfer, so the default must
  be off, the interval conservative, and the cost documented next to the knob. This is the reason it is
  opt-in rather than free.
- **The sentinel write takes the write lock.** litestream-ruby had to add a `busy_timeout` to `verify!`
  for exactly this. Ours should go through `ExecuteInImmediateTransactionAsync` so it waits without
  blocking a thread, and stay tiny.
- **Replication lag is not failure.** "The sentinel had not shipped yet" must be a distinct
  *inconclusive* outcome from "the restore did not contain it" — otherwise the job pages someone every
  time it races the replication interval.
- **Never take the app down.** Follow the existing rule: a verification failure logs `Critical` and
  updates status; it does not throw into the host.
- **Temp files.** Restore to a temp directory, not beside the live database, so a stray `-wal` / `-shm`
  cannot be confused with the real thing; clean up on every path including failure.
- **Config mode is multi-database.** `RestoreAsync` already skips automatic restore when `DatabasePath`
  is empty. Verification has the same ambiguity and should behave consistently rather than guessing.

## Notes

Found while comparing Rask's production-SQLite story against Rails 8 and Laravel 12. Nothing here is a
defect in shipped behaviour — replication works and its liveness is honestly reported. The gap is that
"the replicator is running" is not "the backup is restorable", and only one of those is worth an alert.

Docs to update if this lands: `docs/sqlite.md` (the "Checking that backups are running" section becomes
"...and that they restore"), `llms.txt`, `src/Rask.SQLite.Litestream/NUGET.md`, CHANGELOG.

pal-tamas:
Closes #751.

## The gap

`LitestreamStatus` could report that the replicator was alive. It could not report that what had been
replicated could be read back — and those are different facts. A replica writing to the wrong prefix, a
bucket whose credentials were rotated to read-only, or a `-config` file naming a database nobody writes
to any more all keep `IsReplicating` true and `RestartCount` flat, right up until the restore.

## What this adds

An opt-in verification pass on the existing seams: upsert a sentinel row into the live database, wait for
replication to carry it, restore **to a temp path** that is deleted on every path, and check the sentinel
came back. It publishes to `LitestreamStatus.Verification` as its own record, leaving replication
liveness untouched.

```csharp
o.Verification.Enabled = true;                      // off by default — a pass costs a real restore
o.Verification.Interval = TimeSpan.FromHours(24);   // a daily audit, not a health poll
```

- `ISqliteBackupVerifier` / `LitestreamVerifier`, driven through `ILitestreamExecutor` so the whole pass
  is testable with the existing fake and no binary.
- A `PeriodicTimer` hosted service, registered **only** when `Enabled` — the verifier itself is always
  registered, for a pass on demand.
- **Three outcomes, not two.** `Inconclusive` means the sentinel had not shipped yet — lag, not a broken
  backup — and stays distinct from `Failed`. `LastVerifiedAt` survives an inconclusive pass, so the thing
  to alert on is that timestamp ageing rather than any single outcome.
- The sentinel write goes through Rask.SQLite's non-blocking busy-retry, so it waits out a busy writer by
  yielding the thread rather than holding one, and the fixed-token upsert stays correct if a contended
  `COMMIT` makes it re-run.

## The traps that turned out to be real

- **`-o` in `-config` mode.** `Restore` emitted no output path there at all, so a verification restore
  would have overwritten the live database with a copy of itself. It now takes an output path and emits
  `-o` in both modes.
- **`-if-replica-exists` must be dropped for verification.** It turns "there is no replica at all" into a
  silent success — precisely the failure being hunted. (Not named in the issue; found while building.)
- **litestream refuses to restore over an existing output file**, so each poll attempt deletes the
  previous one first. Confirmed against the binary's own `-force` flag documentation.

## Testing

Unit: 56 tests in `Rask.SQLite.Litestream.Tests` (round trip, three-valued outcome, temp cleanup on every
path, one probe row however often it runs, arg shapes), 4 new dashboard tests, 3 Shop probe tests. Format
+ full unit gate green on both commits; browser E2E green (60/60).

**Validated against a real object store, not only a fake** — committed as
`scripts/verify-litestream-minio.sh` (Docker + MinIO, ~30s, not part of the unit loop):

| Scenario | Result |
| --- | --- |
| A — replication running, replica current | `Verified`, 3.1s round trip, live DB untouched, temp dir empty |
| B — budget too short for the sentinel to ship | `Inconclusive`, `LastVerifiedAt` preserved |
| C — **replica destroyed, replicator alive** | `IsReplicating=True` **and** `Failed` |

C is the thesis of the issue reproduced on real infrastructure. The script also proves the arg shape the
unit suite cannot: the real binary accepts `-o` alongside `-config`.

## Dashboard

The Backup card gains a "Last verified restore" tile beside the replication one, with three levels so lag
reads amber and only a genuinely broken restore reads red. `IDashboardBackupProbe.VerificationAsync` is a
default interface member returning `null`, so existing probes keep compiling; `null` renders nothing at
all rather than good news, since "nobody has checked" is the normal state.

## Note on the dependency

The package now project-references `Rask.SQLite` for the busy-retry and the SQLite provider, so it is no
longer "standalone, only the hosting/DI abstractions". The package description and README no longer claim
otherwise. No benchmarks: nothing here is on the render hot path.


pal-tamas:
Rebased onto current `main` (CHANGELOG conflict resolved keeping both sides).

One thing surfaced by the rebase and worth flagging: **`main` did not compile.** #754 added `ChromeScreen` using `override Route`/`Parent`; #756 removed both along with the `Page` base class. Each was green on its own base, neither rebuilt after the other landed, so `Rask.Example.Shared` — referenced by all three showcase hosts — failed with CS0115 and took the E2E gate with it. Filed as #760.

The two-line migration is carried here as a separate commit (`9d7d73a9`), because this branch's gate could not run until main compiled. It is unrelated to the Litestream work and can be split into its own PR if you would rather land it independently.

Gates on the rebased tree: format + unit green (all suites), browser E2E 61/61, CLI build gate 24/24, plus `scripts/verify-litestream-minio.sh` green against real MinIO.

PR #816, opened 2026-08-25
feat(native): render a hosted app's bars as real platform chrome (#779)

### LINKED ISSUE #779: feat(native): chrome descriptor over the bridge — platform bars in every model
Part of the four-models epic. Mechanism **B**. Closes **G2**.

`NativeChromeDescriptor` is produced only by the in-process session and handed to `INativeChrome`. A
server or wasm page has no way to emit it, so `AppBar`/`TabStrip` render landmark HTML inside the WebView
instead of a real `UINavigationBar` — the biggest "this is a web app" tell.

### Scope

- A handshake flag tells a page it is inside a Rask shell.
- When set, the portable bars emit **no markup at all** and post a `NativeChromeDescriptor` over the
  bridge; the head applies it through the *existing* `INativeChrome` path. Without the handshake nothing
  changes, so the same deployed app keeps its landmarks in an ordinary browser.
- The descriptor, its serializer and both platform bar builders already exist and are tested — this adds a
  second producer, not a new subsystem.
- `PortableChromeTests` / `ScreenChromeTests` extend to the bridged path.

### Done when

One `Screen` subclass produces the same platform bar in all four models, with zero bar markup in the
document, asserted on device.

pal-tamas:
Closes #779.

A `Screen`'s `AppBar` and `TabStrip` became real platform bars only when the app ran **in the head's own
process**. A Server or WASM app inside the same shell drew HTML bars inside the WebView — the same
components, the same `Screen`, and a visibly worse app for no reason a user could see. They cross the
bridge now, so one `Screen` is native chrome in all four models.

## How it works

1. **The head says it is a shell.** `X-Rask-Shell: native` on the load it initiates. A request header, not
   a WebSocket hello: a server renders its document on the GET, so a hello-time handshake would ship bar
   markup that paints and then vanishes. Both platforms also re-issue a same-origin document load that
   lost the header (a reload, a redirect), so a session cannot silently revert to HTML bars mid-flight.
   A WASM app has no request to read — it boots after the document — so the injected bridge sets
   `window.__raskShell` and the session reads it at startup.
2. **The app describes its bars instead of drawing them.** Zero bar markup is emitted. The descriptor
   rides the existing `jsInvokes` channel to `__raskNative.applyChrome`, and the head applies it through
   the same `INativeChrome` the in-process model has always used.
3. **A press still runs your `OnClick`.** The callback lives in the app's C#, wherever that is running,
   so the tap travels back and dispatches like any other event — under the session lock, behind the
   route's authorization check, with `Navigator.NavigateTo` and its history entry working as they do from
   a button in the page.

Only the **portable** bars cross. `NativeHeaderBar`/`NativeToolbar` live in Rask.Native, which a server
app cannot reference — which is exactly why `AppBar`/`TabStrip` exist. The descriptor moved to
Rask.Chrome, the one assembly all three hosts share, and `HostedChromePusher` gives Server and WASM
identical collect/describe/push logic rather than two copies that drift.

## Two defects found on the way

**The first descriptor was stranded.** A server renders its document on the HTTP GET, but `jsInvokes` only
ride WebSocket frames — so the bars described during that render queue into a transport that does not
exist yet, and the app boots with **no bars at all** until the user happens to change something. It
re-pushes at the handshake now. Proved by disabling it: three of four delivery tests fail.

**A tap on a remote bar went nowhere.** The host wired `INativeChrome` to the message router *without* an
evaluator, so the branch forwarding a press back to the page was unreachable. The bar drew perfectly and
did nothing when pressed, while every unit test around it passed — the broken link was the wiring, not
any piece it connects. Found on an emulator. The regression test presses through `OnChromeEvent` on a
really-wired host, which is what the earlier tests could not do.

## Verification

- **iOS simulator**: a genuine `UINavigationBar` (title + trailing button) and `UITabBar` around a page
  whose document contains no bar markup — with a head that declares no bars, so every pixel of that
  chrome came from the server app.
- **Android emulator**: tapped the platform bar's REFRESH; the server's `OnClick` ran and the page went
  from "Refreshed 0 time(s)" to "1".
- 18 new tests: the server WS transport (delivery, taps, unknown ids, the web control), the WASM session
  (shell flag, case-insensitivity, the browser control), and the head (apply, claim, malformed input, tap
  escaping). `dotnet format` + `-warnaserror` + the full unit suite green.

## The E2E gate

The browser E2E gate fails on five WASM samples (Jobs, Playground, Site, both WASM auth journeys) — every
one stuck on its static boot screen. **This reproduces on `origin/main`**: a clean `git archive` of main,
published in a scratch directory with no contamination from this branch, hangs identically. It is also
the failure mode `scripts/run-e2e-local.sh` documents at length as open and intermittent (the bake writing
zero files after a no-native solution build). Removing this branch's only browser-only new code path
changes nothing. Filed separately; this branch is not the cause, and every Server E2E journey passes.


pal-tamas:
Closing unmerged: Rask is going web-only and the native hosting model is being removed, so this feature
has nowhere to land. The branch `feat/chrome-over-bridge` is left on the remote if any of it is ever
wanted back.

For the record, since it affects a claim I made in this PR's description: the E2E-gate justification here
was **not sound**. I asserted the WASM failures reproduce on `main`, based on a throwaway static file
server that turns out to stall part-way through serving the runtime assemblies on its own. That makes the
"reproduces on main" evidence worthless in both directions — it neither convicts nor clears this branch.
See #817, which I have corrected accordingly.

Issue #868, opened 2026-08-27
Nothing scaffolds remote CQRS dispatch since wasm-hosted was removed


pal-tamas:
Removing the `wasm-hosted` template took the only scaffold for remote CQRS dispatch with it.

`Rask.Cqrs.Client` / `Rask.Cqrs.Server` are unchanged and still work, but nothing generates the
arrangement now — an app that wants it wires both packages, `AddRaskCqrsClient()`,
`AddRaskCqrsServer()` and `MapRaskCqrs()` by hand.

## Why this matters more than it did before

It is not simply a lost convenience. `rask new --wasm` produces exactly the app that needs it.

The browser rung moves an eligible page into WebAssembly on the next navigation. A page is *eligible*
when nothing it injects is server-only — and `RASK054` steers people towards reaching data through a
query or a CQRS message precisely because those already cross the wire. But they only cross the wire
when the transport packages are wired. Without them `Rask.Cqrs` dispatches in-process, so a page that
looks eligible either fails in the browser or silently answers from nowhere.

So the shape the one-project build most wants is the shape nothing scaffolds.

## What it probably looks like

Simpler than the old three-project version, because there is no longer a Shared project to put
messages in — `--wasm` compiles one set of sources into both halves, so the messages are already
shared by construction. Handlers stay out of the browser half by living under `Server/`, which is the
convention the companion build already excludes.

Sketch:

- `--wasm` (or `--wasm --cqrs`) adds `Rask.Cqrs.Client` to the companion project and
  `Rask.Cqrs.Server` to the server half.
- `AddRaskCqrsClient()` in the generated browser entry point; `AddRaskCqrsServer()` + `MapRaskCqrs()`
  in `Program.cs`.
- Without `--auth`, `RequireAuthenticatedUser = false` with the comment explaining why — there is no
  authentication to require, and the secure default would make every message answer 401. That
  reasoning came from the removed template and should survive it.

## Notes

- `[LocalOnly]` remains the only way to keep a message in-process, and a client is a pure client —
  worth restating in whatever docs this lands with.
- `docs/cqrs.md`, `llms.txt` and the CHANGELOG currently say plainly that no template scaffolds this;
  those should be updated together with the fix.

PR #873, opened 2026-08-27
feat(cli)!: remove the wasm-hosted template, leaving server and wasm

pal-tamas:
`rask new --template wasm-hosted` is now a usage error naming the templates that remain. Two are left,
and they answer the only question that was ever the author's: **does this app have a backend?**

It scaffolded a Client/Server/Shared trio, which is the hand-written version of what the one-project
build does by itself — except a hosted-WASM app shows nothing until its bundle lands, while
`rask new --wasm` is useful on first paint and degrades to the server if the bundle never arrives.
Pre-1.0, so it is removed rather than deprecated.

## Existing solutions of that shape keep working

Deliberately, and this is the part worth reviewing. Only the *scaffolding* is gone:

- `rask dev` still resolves the `.Server` host
- `rask deploy` still ships it
- the operator dashboard still mounts there

That detection is generic — and this repo's own `samples/Rask.Example.Wasm.Host` has the shape, so
removing it would have broken our own E2E.

## The deleted file was not all template code

`ProjectGenerator.Spa.cs` used one helper from it that re-homes the `DbContext` into a `.Server`
namespace — something every template that splits a client from a host needs, the front-end ones
included. It moved next to its remaining caller and lost the misleading name.

## Tests were changed, not deleted

Nine call sites failed to compile. **Eleven more failed at runtime** — and that is the dangerous half:
they compile fine, because they assert the template's *presence*.

Where the assertion was meaningful it was inverted rather than dropped. The catalog's
`The_native_template_is_gone` and the end-to-end `The_removed_template_is_a_usage_error_and_scaffolds_nothing`
are now theories over **both** removed keys. That is the guard that matters here: the recorded shape of
this bug in this repo is a deletion that leaves a live catalog entry and a default arm that scaffolds
something else entirely and exits 0.

## One capability lost its only scaffold

Remote CQRS dispatch (`Rask.Cqrs.Client` / `Rask.Cqrs.Server`) was wired by this template and by nothing
else. The packages are unchanged and can still be wired by hand, but nothing generates the arrangement —
and a `--wasm` app whose pages move to the browser is precisely what wants it.

Called out in the CHANGELOG, `docs/cqrs.md` and `llms.txt` rather than quietly dropped, and tracked in
**#868**, which also notes the fix should be *simpler* than the removed template's: there is no Shared
project to put messages in any more, because `--wasm` compiles one set of sources into both halves.


pal-tamas:
Superseded by #877 — same content, rebased onto current main after #871 conflicted with it in llms.txt. Closing rather than force-pushing this branch.

Issue #874, opened 2026-08-28
Tests racing on the process-global RaskDiagnostics.Sink fail open


pal-tamas:
`RenderModeAttributeTests.APageDeclaringStaticThatNeedsAConnection_KeepsItAndSaysSo` fails
intermittently. Measured on **pristine `main`** (no local changes): **1 of 8** runs of
`tests/Rask.Server.Tests` alone.

```
Assert.Contains() Failure: Filter not matched in collection
Collection: []
```

The collection is **empty**, not slow — which is the tell. Nothing arrived at all.

## Cause

`RaskDiagnostics.Sink` is process-global, and the test swaps it for a capturing queue, then restores
the previous value in a `finally`:

```csharp
var previous = RaskDiagnostics.Sink;
RaskDiagnostics.Sink = captured.Enqueue;
try { ... } finally { RaskDiagnostics.Sink = previous; }
```

xUnit runs test classes in parallel, and three classes in `Rask.Server.Tests` alone do this
(`RenderModeAttributeTests`, `StaticPageAuditTests`, `RaskServerDiagnosticsBridgeTests`), plus four in
`Rask.Core.Tests`. When two overlap, the second's assignment replaces the first's sink, and the
first's `finally` then restores a value captured *before* any of it — so the diagnostic this test is
waiting for is delivered to a stranger's queue, or to nobody.

It is a save/restore of shared mutable state with no mutual exclusion. The product code is fine; the
harness is what races.

## Why it matters beyond the flake

It fails **open**. A test asserting "the framework reported X" passes whenever the sink happens not to
be stolen, and the same race can make a genuinely missing diagnostic look present in another run —
`StaticPageAuditTests` exists specifically to prove under-detection gets reported, and it captures the
sink the same way.

## Fix shape

Put every class that swaps the sink into one xUnit collection, so they serialise against each other —
this repo already uses that pattern for other process-wide state (`[Collection("WasmSession")]` for
the JSInterop statics). A shared fixture that owns the swap would be tidier still.

## Not this

Unrelated to #870, which is a different intermittent in the same assembly with a different mechanism
(quiescence scope resolution) and a different symptom (a served placeholder).

PR #875, opened 2026-08-28
fix(core): let the running flow's quiescence scope beat a live one left on the thread

### LINKED ISSUE #870: Quiescence still intermittently serves a placeholder: the #865 fix was necessary, not sufficient
`QuiescentRenderTests.Get_AwaitsWorkStartedByAResolvedWave` fails intermittently in a full-suite run
and passes every time in isolation. Roughly 1 in 6 full-gate runs on this machine.

When it fails, the served page contains `child-loading` — the child component's `OnMountAsync` was
never waited for, so the GET shipped a placeholder for data the render should have settled.

## Why this is not the bug that was already fixed

#865 fixed a real leak with the same symptom: `QuiescenceScope.Dispose` clears its thread-static slot
only on the thread it happens to run on, which after an `await` is routinely not the thread `Begin`
ran on — so a finished scope stayed visible to the next render on that pool thread, which then tracked
its work into a dead scope. `Current` now refuses a disposed scope and clears the stale slot, and
`QuiescenceScopeStaleThreadTests` pins it.

That fix is in and the symptom still recurs. It was **necessary and not sufficient.**

## Hypothesis (unverified)

A race between mounting a child and registering its work.

The wave loop exits when `TrySnapshotPending` reports nothing pending. The fixture's child only
exists once the parent's data lands, so its `OnMountAsync` starts during wave 2. If that hook is
*dispatched* rather than run inline — through the lifecycle sync context — the render can return, the
loop can snapshot an empty pending set, and the loop can exit before the child's task is ever
registered. Under load the dispatch is likelier to be late, which matches "passes alone, fails in the
suite".

If that is what is happening, the fix is not in the loop: registration has to happen before the
render that mounted the component is observable as finished.

## Where it lives now

The loop moved to `Rask.Core.Live.QuiescentRender` (host-agnostic, so a build-time prerender can use
it). The move is behaviour-preserving and did not cause this — the assembly passes 391/391 three runs
in a row after it, and the failure predates it.

This matters more than a flaky test: the same race in production serves a placeholder and answers
`200`, which is the SSR path failing silently rather than loudly.

## Reproducing

`./scripts/run-unit-local.sh` repeatedly. It is load-dependent; a single assembly run
(`dotnet test tests/Rask.Server.Tests`) has not reproduced it in 3 attempts, the full gate has once in 6.

pal-tamas:
Closes #870.

`QuiescenceScope.Current` consulted the thread-static slot before the flow's `AsyncLocal`. A pool
thread keeps whatever scope `Begin` set on it — only a `Dispose` that happens to run on *that same
thread* clears it — so a thread can be carrying **another render's live scope**. When a render's later
wave resumed there, the work it started was tracked against a stranger, its own wave loop saw nothing
pending, and the page went out with a placeholder for data it never waited for.

It answers `200` while doing it, so nothing reports a fault. In the suite it looked like a flaky test;
in production it is the SSR path failing silently.

## The hypothesis in the issue was wrong

Registration is not late. `Track` runs **synchronously inside the render walk**, immediately after the
hook returns. The question was never *when* the work is registered but *which scope receives it*.

## Why flow-first is safe

Nothing needed the thread to win. The one path that loses the `AsyncLocal` —
`LifecycleSyncContext`'s suppressed `Task.Run` — restores the captured scope with `Enter`, which sets
**both** slots, so a flow-first lookup finds it there too.

`LiveRenderContext` already resolves this way: its `Current` is `AsyncLocal`-only, with the
thread-static a separately named accessor guarded by `IsActive`. `QuiescenceScope` had simply diverged
from its sibling.

## Supersedes the earlier fix

#865 fixed the same symptom by skipping *disposed* foreign scopes. That was necessary and not
sufficient: it covered the dead case, this covers the live one.

## Evidence, stated honestly

Eight full-gate runs are clean, against a prior rate of roughly one in six. That is **support, not
proof** — eight clean runs would happen by luck about a quarter of the time at that rate. The weight
sits on the mechanism being named and the rule being pinned by tests, mutation-checked by reversing
the precedence.

Two earlier verification attempts proved nothing, and are recorded in the commit so they are not
repeated:

- **The server suite alone has never reproduced this.** Only the full-solution gate has — the symptom
  needs that parallel load. Runs against the suite alone were measuring nothing.
- **A grep for the test name matches the passing lines too.** The first detector reported 8/8
  "failures" that were all passes.

## Also found, filed separately

`RenderModeAttributeTests.APageDeclaringStaticThatNeedsAConnection_KeepsItAndSaysSo` is a *different*
intermittent in the same assembly — measured at 1 of 8 on pristine `main`, so pre-existing and
unrelated. Seven test classes race on the process-global `RaskDiagnostics.Sink`, and it fails open.
See #874.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,27 @@ them until tagged releases begin.
 ## [Unreleased]
 
 ### Fixed
+
+- **A render could hand its async work to a different render, and then serve a placeholder for it.**
+  `QuiescenceScope.Current` consulted the thread-static slot before the flow's `AsyncLocal`. A pool
+  thread keeps whatever scope `Begin` set on it — only a `Dispose` that happens to run on that same
+  thread clears it — so a thread can carry another render's **live** scope. When a render's later wave
+  resumed on such a thread, the work it started was tracked against a stranger, its own wave loop saw
+  nothing pending, and the page went out with a placeholder for data it never waited for.
+
+  It answers `200` while doing it, so nothing reports a fault. In tests it surfaced as an intermittent
+  failure that passed every time in isolation; in production it is the SSR path failing silently.
+
+  The flow now wins over the thread. Nothing needed the other order: the one path that loses the
+  `AsyncLocal` — `LifecycleSyncContext`'s suppressed `Task.Run` — restores the captured scope with
+  `Enter`, which sets both slots. `LiveRenderContext` already resolved this way; `QuiescenceScope` had
+  diverged from it.
+
+  Supersedes the earlier fix for the same symptom, which skipped *disposed* foreign scopes. That was
+  necessary and not sufficient — it covered the dead case, this covers the live one.
+
+
+### Fixed
 - **The browser gate no longer needs PowerShell, and no longer skips itself quietly when it is
   missing.** `scripts/run-e2e-local.sh` installed the Playwright browsers by shelling out to
   `pwsh <path>/playwright.ps1 install chromium`, and skipped that step whenever `pwsh` was absent.
--- a/src/Rask.Core/Live/QuiescenceScope.cs
+++ b/src/Rask.Core/Live/QuiescenceScope.cs
@@ -41,34 +41,62 @@ internal sealed class QuiescenceScope : IDisposable
 
     /// <summary>The scope collecting work for the render currently running, if any.</summary>
     /// <remarks>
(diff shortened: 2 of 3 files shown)


pal-tamas:
Superseded by #885 — same content, rebased onto current main after a CHANGELOG conflict. Closing rather than force-pushing this branch.

PR #876, opened 2026-08-28
fix(server): make RenderModes.ServerInteractivity do what its name says

pal-tamas:
`RenderModes.ServerInteractivity` was read **nowhere outside its own validation**. Turning it off with
`Wasm` on was accepted and silently ignored — pages still opened WebSockets — and turning it off with
`Wasm` off threw, telling you to turn it back on. It lied in both directions.

It now means one thing: **no page ever gets a live session.** Every page is served as a document, and
the socket endpoint answers `404` as though it were not there.

## Why not just use `RenderModes.Static`

That was the old error message's advice, and it is backwards. `Static` is **detected** per page and
deliberately biased towards keeping a connection — a false `interactive` is a non-event, a false
`static` is silent breakage. So a content page that happens to render one handler quietly takes a
session and a socket.

`ServerInteractivity = false` is **declared**. Nothing is detected, so nothing can bias. Turning
`Static` on as well is *not* required — a test pins that, because requiring it would make the weaker
statement load-bearing.

## Both combinations now work

| | |
|---|---|
| `ServerInteractivity = false`, `Wasm = false` | Plain server-side rendering. A content site. Used to throw. |
| `ServerInteractivity = false`, `Wasm = true` | Static HTML handing over to WebAssembly, no socket ever opened — the edge-hosted arrangement from the plan, previously unreachable because the switch did nothing. |

## Implementation note worth reviewing

The gate wraps the **whole** interactivity expression rather than joining it as another clause:

```csharp
var interactive = limits.ServerInteractivity
                  && (!(staticPages || declaredStatic)
                      || session.RequiresLiveSession
                      || ...
                      || LiveOptions.IsDevelopment == true);
```

Every clause inside assumes a session is *available*, and none can be honoured here — including the
`IsDevelopment` one, which would otherwise make every page live while you develop the very thing you
turned off.

`404` rather than `400`/`426` on the socket, deliberately: an app that cannot go live should be
indistinguishable from one that never had a socket.

## Tests

Two existing tests asserted the old refusal; both are **inverted rather than deleted**, and the
end-to-end one now uses a contradiction that still is one (the browser rung on with nowhere to fetch
the bundle from).

The new cases use a page that renders a **button** on purpose. A page needing nothing live proves very
little here, since `Static` already serves those as documents — what has to hold is that a page which
*would* have taken a session does not get one.

Mutation-checked: removing the gate fails three of the six.

## The cost, reported not prevented

A page that renders a handler in this mode has nothing to answer it. That goes out through the
`Rask.Ssr` diagnostic. Refusing to start was the wrong response, because it made "serve only content"
unreachable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,27 @@ them until tagged releases begin.
 ## [Unreleased]
 
 ### Fixed
+
+- **`RenderModes.ServerInteractivity` did nothing.** The property was read nowhere outside its own
+  validation, so turning it off with `Wasm` on was accepted and silently ignored — pages still opened
+  WebSockets — and turning it off with `Wasm` off threw, telling you to turn it back on. It lied either
+  way.
+
+  It now means what it says: **no page ever gets a live session.** Every page is served as a document,
+  and the WebSocket endpoint answers `404` as though it were not there. That is stronger than
+  `RenderModes.Static`, which is detected per page and biased towards keeping a connection; this is
+  declared, so nothing is detected and nothing can bias.
+
+  Both combinations now work rather than one throwing: with `Wasm` off it is plain server-side
+  rendering — a content site, which used to be refused — and with `Wasm` on it is static HTML that
+  hands over to WebAssembly with no socket ever opened.
+
+  A page that renders a handler in this mode has nothing to answer it. That is reported through the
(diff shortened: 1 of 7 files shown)


pal-tamas:
Superseded by #882 — same content, rebased onto current main after a CHANGELOG conflict. Closing rather than force-pushing this branch.

PR #458, opened 2026-07-16
feat(cli): rich per-command --help with options table, examples, and color

pal-tamas:
## Summary
`rask <command> --help` was three lines (summary + one usage string). It now renders a described **Options** table straight from the same `ArgumentSchema` that parses the arguments (so help can't drift), an **Arguments** section, and copy-pasteable **Examples**. This surfaces `rask generate`'s previously *undiscoverable* feature flags — `--bs`, `--modal`, `--soft-delete`, `--concurrency`, `--events`, `--outbox`, `--tests`, `--validation`, `--no-restore` — grouped under "Feature options". Output is colorized on a terminal and stays plain when piped or under `NO_COLOR` (so `rask info | cat` and CI logs are unaffected); parse errors now point at `rask <command> --help`. Pure BCL — no new dependencies.

This is the first of four sequenced CLI-DX PRs (rich help → colored/consistent output → interactive prompts/wizard → completion + config defaults). It lands the shared foundation: the styling/redirection seam on `IConsole` and the self-documenting `ArgumentSchema`.

## Testing
- Unit: `dotnet test Rask.slnx --filter "FullyQualifiedName!~Rask.Examples.E2E"` — **250+335 pass, 0 fail** (full local gate green: whole-solution Release build, whitespace format check, all non-E2E tests). 10 new CLI tests (`CommandHelpTests`, `ConsoleStylingTests`, `ArgumentSchema.Declared`).
- CLI smoke: `.claude/skills/run-rask-cli/smoke.sh` — **15/15 pass**, incl. new checks that `generate --help` surfaces `--outbox`/`--tests` and an examples section.
- E2E: n/a — CLI-only change, no `samples/` or render-path code touched (pre-push E2E gate skipped per its documented bypass for non-UI pushes).

## Benchmarks
n/a — the CLI is not on the render hot path.

## CHANGELOG
- [Unreleased] entry added: `rask <command> --help` now teaches — aligned options table, arguments, examples, color, and the surfaced generate feature flags.

PR #880, opened 2026-08-28
fix: move the bootstrap traps out of scaffolded comments and into the framework

pal-tamas:
## Summary

Phase A of moving Rask's bootstrap toward convention: **the framework now decides the things a scaffolded
`Program.cs` used to ask you to remember.**

Four of the comments in the generated `Program.cs` were not documentation — they were hand-avoidance
instructions for silent failures. Each one only ever reached apps generated *after* the comment existed; an
app written by hand, or one scaffolded a year earlier, got the bug with nothing in the logs. This moves each
into the framework, where it holds for every app.

| Was | Now |
| --- | --- |
| `AddRaskData(o => o.DispatchDomainEventsInProcess = false)` — forget it and the outbox stays empty | `AddRaskOutbox` claims delivery; `AddRaskData()` takes no argument, in any order |
| 16 lines of `PersistKeysToFileSystem` — without them every deploy signs all users out | `AddRask` persists the key ring when the host has somewhere durable |
| `Configure<HostOptions>` — without it the pillars' shutdown graces sum past the SIGKILL | `AddRask` sets the budget and stops services concurrently |
| A comment warning that a second `AddRask` is silently dropped | **RASK056** reports it in your own code |

### The bugs, concretely

**Domain events vs. the outbox.** `AddRaskData` decided whether to register the in-process publisher at the
moment it was called, from an option you had to set by hand. Forget it — or write the two `Add` calls in the
order `AddRaskData` first — and both interceptors ran. `DomainEventInterceptor` drains and **clears** every
entity's events in `SavingChanges`, before `OutboxInterceptor` can copy them: the outbox table stayed
permanently empty, delivery quietly stopped being durable, and **nothing failed**, because the handlers still
ran in-process. The decision now happens when the container is built, from whether anything registered an
`IDomainEventDeliveryOwner`.

**The key ring.** The default ring lives inside the container and every deploy replaces the container, so
everything sealed under the old one silently stops opening — every auth cookie already issued, and every Rask
session-resume record. `Rask.Server` already *depended* on a persisted ring for session resume while leaving
the fix to scaffolded user code.

**The shutdown budget.** Stopped one at a time (.NET's default) the pillars' graces sum to 30s against a
20s SIGKILL, and whichever service stops last gets none — decided by the order of your `AddRaskX` calls, so
the symptom is a truncated write that moves when you reorder two unrelated lines. `SamplesShutdownBudgetTests`
exists because **nine of the ten web-host samples in this repo** were sitting on the wrong default.

### Also corrected

The scaffolder claimed `AddRaskOutbox` must precede `AddDbContextFactory` "so its interceptor is in the
container when the factory resolves `ISaveChangesInterceptor`". It does not — that callback runs when the
factory is first resolved, after `Build()`, so it sees every registration whenever it was made. Now covered
by a test instead of a comment.

### What review caught, and it was load-bearing

The first pass put the defaults in `Rask.Server`'s `AddRask` and removed the hand-written blocks from
**every** template. But the WASM bundle host calls `Rask.Wasm.Hosting`'s own `AddRask`, the SPA host calls
`AddRaskSpaHost`, and neither did anything beyond response compression — while the `wasm-hosted` template
only reaches `AddRaskServer` when the operator dashboard is on, and the `spa` template called no host
registration at all.

So a cookie-authenticated `rask new --template wasm-hosted` without `--ops` would have shipped with **no
persisted key ring**: signed out on every deploy, the exact bug this set out to fix. Every host-level unit
test stayed green, because they all called the one `AddRask` that had been fixed.

Both setups now live in `src/Rask.Hosting.Shared` — which the two hosts already source-link — and all three
hosts register them. `HostDefaultsTests` in each host's test project asserts the seam rather than the part.
The `spa` template also gains `AddRaskSpaHost()`, which it should always have had: it was scaffolding
`UseRaskSpa()` with no matching `Add`, so it shipped without compression too.

Three more findings, each verified before fixing:

- **`ShutdownTimeout` silently discarded `ASPNETCORE_SHUTDOWNTIMEOUTSECONDS`** and the
  `shutdownTimeoutSeconds` key — the generic host binds those *before* `AddRask` runs. It is now only
  replaced while it still holds .NET's 30s default. `ServicesStopConcurrently` stays unconditional: it has
  no configuration binding to defer to.
- **`ApplicationDiscriminator` was being reverted.** ASP.NET's own `DataProtectionOptionsSetup` writes it
  without checking, so when Data Protection arrived later — pulled in by the scaffold's `AddAuthentication`
  — the ring persisted but two containers still derived different keys. `AddDataProtection()` now runs
  first, unconditionally. The existing tests could not see this, because they all called it first themselves.
- **`Directory.CreateDirectory` could surface as a 500.** It runs from a lazily-resolved options setup, so
  an unwritable `/data` — a conventional mount point, root-owned against a non-root container — failed on
  the first cookie operation rather than at boot. It now falls back to the framework ring and logs the fix.

RASK056 gained two exclusions it was reporting wrongly (a constructed receiver; calls on opposite branches
of an `if`/`else`), and its message no longer claims the later call is discarded — culture options keep the
**first** call's and the rest take the **last**, so which wins depends on the option.

## Testing

- `scripts/run-unit-local.sh` — **48 assemblies, 6966 cases, 0 failures**, and deterministic: two
  consecutive runs agree, and the total reconciles exactly to the pre-change baseline plus the tests added
  here
- `dotnet build Rask.slnx -warnaserror` clean; `dotnet format --verify-no-changes` clean
- `scripts/run-e2e-local.sh`

**Verified against the previous behaviour rather than assumed.** With the framework side reverted to base,
`OutboxDeliveryHandoverTests.A_plain_AddRaskData_hands_delivery_to_the_outbox` fails and
`The_handover_holds_when_the_outbox_is_registered_first` passes — which is the order-dependence itself, not a
flaky test.

Two assertions were corrected after they failed for the wrong reason, both worth recording:

- ASP.NET's *own* development key ring is also a `FileSystemXmlRepository`, so a type check would have passed
  whether or not Rask had overridden it. The test asserts the decision instead.
- `DataProtectionOptions.ApplicationDiscriminator` defaults to the **content root path**, not null — which is
  exactly why pinning it matters when a ring is shared across build and runtime images.

Three CLI tests that asserted the scaffolder *emits* a block now assert its **absence**, with the guarantee
re-pointed at where it actually lives. `SamplesShutdownBudgetTests` stops scanning sample source text and
resolves the options `AddRask` produces; `Rask.Cli.Tests` takes a test-only reference on `Rask.Server` so the
deploy ladder and the framework default stay coupled, since the two numbers live in assemblies that cannot
see each other by design.

## Changelog

`[Unreleased]` — one `Added` entry (RASK056) and three `Fixed`.

## Notes

- **RASK056, not RASK055** — #871 took 055 for the scoped-JavaScript error while this branch was in flight.
- `RaskDataOptions.DispatchDomainEventsInProcess` becomes `bool?`: `null` (default) is automatic, `true`
  forces in-process publishing even alongside an outbox, `false` never registers the interceptor. Existing
  `= false` call sites still compile and keep their meaning — they are simply no longer needed.
- Behaviour change on upgrade, flagged in the CHANGELOG as a fix: an app relying on in-process dispatch
  *alongside* an outbox stops double-dispatching. That combination is the documented bug.
- `Program.cs` in the Shop sample goes 322 → 284 lines. Phase A was about correctness, not compression —
  the facade (`RaskApp.Create`) is what collapses the rest, and is not in this PR.


pal-tamas:
Landed as part of #888, squashed into `b8b05334` — this branch's five commits are contained in it.

Closed rather than merged because this repo lands a stacked series as one squash via the final PR (precedent: #461, with #458–#460 closed).

Two things changed while landing, both worth knowing:

- **RASK056 → RASK060.** `main` gained #864 after this branch was cut, which allocated RASK056–RASK059 to external components. `DuplicateAddRaskAnalyzer` is renumbered to **RASK060** across the descriptor, its `docs/diagnostics.md` section (moved after RASK059), the TOC row, the scaffold comment, the tests and their `#pragma`s. `CLAUDE.md` now records RASK061 as next.
- **`ProjectGenerator.WasmHosted.cs` was deleted, not modified.** #877 removed the wasm-hosted template. This branch's edit to that file was verified purely subtractive (−23, zero additions) and identical to the refactor applied across `.Server.cs`, `.Spa.cs`, `.Shared.cs` and the samples, so the deletion was taken. `ShutdownBudgetBlock` and `GenerateWasmHosted` have zero remaining references.

Nothing else is dropped — the outbox delivery ownership, the Data Protection key ring, the shutdown budget and the three-host `Rask.Hosting.Shared` setup are all on `main` as of that commit.

PR #889, opened 2026-08-28
feat(core): render a whole document with no host, and plan which routes can be

pal-tamas:
The first two pieces of build-time prerendering, both in Core and both usable on their own.

A standalone WASM app ships its boot shell to every visitor and every crawler — a spinner and the word
"Loading" — because the real markup does not exist until several megabytes of runtime have downloaded.
That includes this framework's own docs site on GitHub Pages.

## `RaskPrerender.RenderDocumentAsync`

The app's root, wrapped in the same boundary both hosts install, driven through the same waves a
server's first response uses. A page whose `OnMountAsync` loads build-time data writes **the data**,
not its placeholder — which is the whole reason it goes through the wave loop rather than rendering
once.

It is small because #872 moved that loop into Core first.

**`Faulted` and `TimedOut` are on the result for one reason: both still return perfectly ordinary
HTML.** A root boundary renders an error document; a timed-out page renders its placeholder. A caller
that writes either blindly publishes it under the route's own name and nothing at build time says so.
Baking a spinner is worse than not prerendering the route at all, because it *looks* prerendered. The
two negative tests are the ones that matter here.

It deliberately does **not** take a route — the caller seeds `RouteState`, because the caller holds the
route table.

`RenderAsLiveRoot` and `RootErrorBoundary` stay internal. Widening them would expose `publishOnly`,
whose contract is that honouring it is mandatory or every wave re-fires `OnRendered` on everything the
last one rendered. That is not something a caller should have to understand.

## `RaskPrerender.PlanRoutes`

Splits the registered route table into what a pass can and cannot cover.

**`Skipped` is a field on the result rather than a log line**, and that is the decision worth arguing.
A parameterised route cannot be enumerated without the values and a catch-all is a 404 page at best —
so a pass that quietly dropped them would cover a site's static half and report as though it covered
all of it. Making the caller *receive* them means it has to decide what to say. This repo has met that
shape repeatedly: a gate filter that selected no tests, a battery list covering everything except the
newest flag, a collection containing only itself.

Prerenderability is decided on the **parsed segment kinds**, not by looking for a brace in the
template. Both agree today; only one keeps agreeing if the syntax ever grows a construct carrying no
brace, or a literal that contains one.

`RouteFlattener` and `RoutePattern` stay internal — the plan lives beside the renderer and exposes only
its result.

## Not in this PR

The build integration: a companion project compiled from the app's own sources for a desktop TFM, run
at publish, walking the plan and writing `index.html` per route. That is where the MSBuild traps live,
and it is worth landing separately.

### DIFF
--- a/src/Rask.Core/Live/RaskPrerender.cs
+++ b/src/Rask.Core/Live/RaskPrerender.cs
@@ -1,3 +1,5 @@
+using Rask.Core.Routing;
+
 namespace Rask.Core.Live;
 
 /// <summary>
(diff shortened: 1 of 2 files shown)


pal-tamas:
Landed as part of #890, squashed into `463dd1c4` — this branch's two commits (`RaskPrerender.RenderDocumentAsync` and `PlanRoutes`) are contained in it verbatim.

Closed rather than merged because this repo lands a stacked series as one squash via the final PR (precedent: #461, with #458–#460 closed). Nothing here is dropped: `src/Rask.Core/Live/RaskPrerender.cs` and `tests/Rask.Core.Tests/Live/RaskPrerenderTests.cs` are on `main` as of that commit.

Issue #886, opened 2026-08-28
CLI build gate: the angular case fails on Node 24.14 (Angular CLI raised its floor), blocking docs-only pushes


pal-tamas:
`SpaTailwindBuildE2ETests.A_utility_class_reaches_the_bundlers_emitted_css(frameworkKey: "angular")`
fails before any Rask code runs:

```
[angular] npx failed.
Node.js version v24.14.0 detected.
The Angular CLI requires a minimum Node.js version of v22.22.3 or v24.15.0 or v26.0.0.
```

The Angular CLI raised its Node floor. `npx @angular/cli new` now refuses on Node 24.14.0, so the
scaffold never happens and the assertion never gets a chance to run. The `react` case in the same
theory passes, as does everything else in the gate.

**Why this is more than one red test.** The CLI build gate is path-gated on
`generator_paths` (`.githooks/pre-push:87`), which includes `src/Rask\.[A-Za-z.]+/` and
`docs/tutorial/`. Those match *any* file under any `src/Rask.*` project — including a markdown file.
I hit this from a branch whose only matching changes were `src/Rask.Cli/NUGET.md` and
`docs/tutorial/00-overview.md`, neither of which can affect scaffolding output. So on a machine whose
Node is behind Angular's floor, **every push touching `src/**` or `docs/tutorial/**` is blocked**,
whatever it changed.

Two separate things worth deciding:

1. **The Node floor.** `ProjectGenerator.Spa.cs:128` tells users "Node.js 22.12 or newer — ideally
   the current LTS". That is no longer sufficient for the `angular` template specifically: Angular
   wants 22.22.3 / 24.15.0 / 26.0.0. Either the message and the docs get a per-template floor, or
   the angular case pins an Angular CLI version whose floor the documented minimum satisfies.
   Pinning is probably the better trade — right now an external project's release schedule can turn
   this gate red on an unchanged repo, which is the same class of problem as an unpinned action tag.

2. **The gate's blast radius.** `src/Rask\.[A-Za-z.]+/` matching a `.md` is deliberate conservatism,
   but it means a docs-only change pays for the full pack-and-build gate and inherits any unrelated
   failure in it. Worth considering whether the regex should exclude `**/*.md`.

Local workaround while the machine's Node is behind: `RASK_SKIP_CLI_BUILD_E2E=1 git push`, having
confirmed the change touches no scaffolding source.

Found while pushing the `rask.sh` installer branch. Related: the installer now provisions Node LTS
[...]

PR #899, opened 2026-08-31
fix(gates): say when a gate is skipped, and name contention when a suite goes red

### LINKED ISSUE #845: A push from a worktree runs the main checkout's .githooks, so a hook change is never tested by its own PR
A push from a `git worktree` runs the **main checkout's** `.githooks/`, not the worktree's. So a change to a hook is never exercised by the push that introduces it — and this repo's whole workflow is worktree-based.

### Mechanism

`core.hooksPath` is set to `.githooks` — a **relative** path. Git resolves a relative `core.hooksPath` against the working tree of the **common** git dir, which for a linked worktree is the main checkout:

```
$ git config core.hooksPath
.githooks
$ git rev-parse --show-toplevel      # the worktree
/Users/…/Rask/.claude/worktrees/wasm-boot-reporting
$ git rev-parse --git-common-dir     # the MAIN repo's .git
/Users/…/Rask/.git
```

So `git push` from that worktree executes `/Users/…/Rask/.githooks/pre-push` — whatever the main checkout happens to have on disk, including uncommitted edits, and regardless of what the branch being pushed contains.

### How it surfaced

A branch added a gate to `.githooks/pre-push` (wiring up the orphaned `scripts/run-wasm-watch-e2e.sh`) and extended `watch_paths` to cover `src/Rask.Wasm/`'s client files. The push touched exactly those files. The gate did not run, and printed nothing — the path filter's silent `:` branch was taken, because the hook that executed was main's, whose `watch_paths` has no `src/Rask.Wasm` entries.

Everything else looked completely normal: the browser gate ran (61/61) and the CLI build gate ran (30/30), because those parts of main's hook are identical to the branch's. Only the *new* gate was missing, and a gate that does not run is indistinguishable from one that passes.

Verified the wiring itself is correct by dry-running the branch's own hook with the gate runner stubbed — it selects all four gates. It simply was not the file git ran.

### Why it matters beyond that one gate

- **Every hook change in this repo is unverifiable by its own PR.** You cannot test a `pre-push` change by pushing, which is the only way anyone would think to test it.
- **The guarantee is weaker than documented.** `CONTRIBUTING.md` and `docs/development-workflow.md` say the gates are "enforced by `.githooks/pre-push`". For worktree-based work that is really "enforced by whatever `.githooks` the main checkout currently has" — which may be an older commit, or dirty.
- It compounds the failure mode already catalogued here: a gate that silently does not run. This is a fifth way, and unlike the others it is invisible from inside the branch.

### Options

1. **Make `core.hooksPath` absolute per worktree.** Correct per checkout, but it is repo-level config shared by all worktrees, so it cannot differ between them — this does not actually work.
2. **Have the hook re-exec the version from the pushed worktree.** A thin `/…/Rask/.githooks/pre-push` that locates the pushing worktree's top level and delegates to *its* `.githooks/pre-push`. Fixes it for every future hook change, at the cost of the shim itself never being testable that way.
3. **Assert it.** Have the hook compare itself against the pushing worktree's copy and fail loudly when they differ, so the discrepancy is named rather than silent. Cheapest, and it converts a silent miss into a message.
4. **Document it** and accept that hook changes need a manual run. Weakest, but honest.

I lean 3 + 4: a loud mismatch message costs nothing and turns an invisible failure into a visible one, and the manual-run instruction is needed regardless while the shim itself is unverifiable.

Found while wiring the WASM watch gate; that PR states plainly that its new gate is dry-run-verified rather than exercised, for this reason.

### LINKED ISSUE #850: The E2E concurrency guard only catches a second browser gate; a build competing with a live suite is still silent
`scripts/run-e2e-local.sh` now refuses to start when another **browser gate** is already running, and names it — pid, elapsed, worktree (landed in #847). That is the right guard for the collision it was built for, and it is not the collision that costs the most.

The expensive case is anything *else* competing with a live browser suite: a `pre-commit` hook, a plain `dotnet build`, a `dotnet publish`, the CLI build gate. None of those are browser gates, so nothing refuses, nothing warns, and the contention is silent.

### Why it matters more than the case that is guarded

The browser journeys and the WebSocket tests are timing-sensitive. Under load they fail in ways that look exactly like real bugs, minutes after the contention, with nothing in the log pointing back at it. Two data points from one afternoon:

1. **A commit whose `pre-commit` hook ran while a browser gate had the machine** failed `HelloMessageTests.Hello_StateMutatedBeforeHello_EmitsCatchUpFrame`, a WebSocket catch-up-frame timing test. It then passed **5/5 in isolation** and **336/336** for the whole assembly immediately afterwards, and the change under test was CLI scaffolding — there is no causal path from it to that test. (Reproduction courtesy of the `batteries included default` session.)

2. **An unexplained five-minute timeout** between two worktrees that were *actively coordinating* about the machine, while a third and then a fourth suite ran unseen. That one is what motivated #847's guard in the first place.

The second is now caught. The first is not, and it is the more common shape — every commit runs a `pre-commit` hook, and nothing stops one landing in the middle of someone's browser suite.

### Why the existing guard cannot be extended as-is

It detects *its own kind*: `pgrep` for another `run-e2e-local.sh`, matched on argv's executable position. Extending it to "is any heavy build running" is a different and much fuzzier question — `dotnet` is running for dozens of legitimate reasons, most of them cheap, and refusing on any of them would block far more than it protects.

### Options, roughly in order of cost

1. **Make the browser gate advertise itself, and have the cheap gates check.** A pid file written by `run-e2e-local.sh` for its lifetime is the obvious mechanism — but a lockfile survives `kill -9` and a laptop sleep, which is precisely why #847's guard uses process detection instead. A process-detected version is the same `pgrep` the guard already has, called from the other side: `pre-commit` could warn (not refuse — a blocked commit is worse than a slow one) when a browser gate is live.

2. **Say it in the failure, not before it.** When a timing-sensitive suite goes red, have the harness check whether a browser gate was running and add one line to the failure: *"a browser E2E gate was live during this run; re-run alone before investigating."* This does not prevent anything, but it attacks the actual cost, which is the hours spent treating a load flake as a real bug. Cheaper than (1) and it cannot block anyone.

3. **Serialise properly** — a real queue rather than refuse-and-retry. Correct, and considerably more machinery than this repo needs for a handful of concurrent sessions.

I lean **2, then 1**: naming the suspicion at the moment of failure is where the money is, and it has no false-positive cost because it adds a hint rather than a decision. (1) is worth having but only warns, since a `pre-commit` that refuses would be a worse tax than the flake.

Worth being explicit that this is a limitation of what #847 shipped, not a regression: before it, *nothing* was guarded and four sessions collided in one afternoon without noticing. This is the next layer, and naming it beats letting the merged guard read as more complete than it is.

pal-tamas:
Closes #845. Closes #850.

Two reports about one failure mode: a gate whose silence cannot be told from success.

## The silent skips (#845)

The four path-filtered `pre-push` gates — CLI build, watch hot-reload, deploy, install — took a bare
`:` branch when nothing in a push matched their paths and printed **nothing at all**. Each now prints
one line naming the filter. Visible in this branch's own push:

```
pre-push: CLI build gate SKIPPED — nothing in this push matches the generator paths.
pre-push: watch hot-reload gate SKIPPED — nothing in this push matches the hot-reload paths.
pre-push: deploy gate SKIPPED — nothing in this push matches the deploy paths.
pre-push: install gate SKIPPED — nothing in this push matches the installer paths.
```

One filter was also wrong: `src/Rask.Core/build/` was missing from `watch_paths`, so a change to
`Rask.Core.targets` — the file that *builds* `@(Watch)`, and therefore decides what `dotnet watch` can
see at all — did not select the hot-reload gate. The one hot-reload change that filter could not catch
was a change to the watch list itself.

## #845's stated mechanism is not the cause

The issue reported that a push from a worktree runs the **main checkout's** hooks, because
`core.hooksPath` is relative. **It does not.** Git resolves that path against the pushing worktree's
own top level.

Verified on git 2.50.1 in a throwaway repo — never this one, since other sessions push here — with a
linked worktree whose `.githooks/pre-push` differed from the main checkout's:

| case | hook that ran |
|---|---|
| push from the worktree root | the **worktree's** |
| push from a subdirectory of the worktree | the **worktree's** |
| worktree branch with no `.githooks/` at all | **none, silently** — no fallback to the main checkout |

So a hook change *is* exercised by the push that introduces it, and the shim the issue proposed is not
needed. The third row is a real caveat that came out of the investigation and is now documented.

`CONTRIBUTING.md` and `docs/development-workflow.md` state the resolved behaviour instead of leaving it
to inference.

## Contention, named at the moment of failure (#850)

The existing guard detects its own kind — a second browser gate — and refuses. The expensive collision
is everything else: a `pre-commit` hook, a plain `dotnet build`, a publish. None of those is a browser
gate, so nothing warned.

- `run-e2e-local.sh`'s test run was the script's **last statement** under `set -e`, so a red suite
  propagated an exit code with no explanation layer — unlike its own build step, which is explained.
  It and `run-unit-local.sh` now report a competing heavy build and ask for a re-run alone first.
- MSBuild worker nodes are ignored, so one build is not reported as eight; the cheap verbs (`run`,
  `ef`, `tool`, `--version`) are not named at all, or the hint fires on every red run and stops being
  read.
- `pre-commit` warns when a browser gate is live and **never refuses** — every commit runs that hook,
  and a blocked commit is worse than a slow one.

These are hints, never decisions. They add a line to a failure that has already happened, so a false
positive costs a sentence rather than a blocked push.

## The refusal was being classified as a test failure

The guard's own refusal fell through to `unknown`, so the hook printed *"look for a failing assertion,
a timeout, or a host that exited early"* — about a suite that never started — directly beneath the
guard's correct *"wait for the run above to finish"*. There is now a `busy` kind.

It sits **below** `code`: a real `error CS` still wins, because if something got far enough to fail
compiling then something did run, and blaming the machine there would be #718 inverted. That ordering
was caught by its own test, which failed first.

The fix validated itself on this branch's first push attempt:

```
pre-push: E2E gate (browser journeys): nothing ran — the machine was already busy with another browser gate.

  The guard above refused to start this suite and named the run that holds the machine.
  There is no failure here to investigate: no journey ran, no assertion was evaluated.
```

## Testing

- Full unit gate green: **6933 cases**, plus the bash suites — `build-failure-kind` 17,
  `e2e-concurrency` 33, `install-script` 88.
- Browser E2E gate green on the push that opened this PR.
- New table cases cover `rask_other_heavy_builds` (real verbs vs. cheap ones, MSBuild worker nodes,
  self, already-exited pids) and the `busy` classification — including "a test that merely quotes the
  refusal is not busy", so asserting on the guard's own wording cannot make every red run look like
  contention.
- The `pre-commit` warning block is exercised against the **shipped** hook rather than a copy: it warns
  on a live gate, stays silent when idle, and stays silent on an editor that merely mentions the
  script.

## Not included

Excluding `**/*.md` from `generator_paths` — the other half of #886's report — is left out
deliberately: it was not asked for, and it touches lines adjacent to these changes.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -46,6 +46,31 @@ fi
 
 root="$(git rev-parse --show-toplevel)"
 
+# WARNS, never refuses (#850). Every commit runs this hook, and nothing stops one landing in the middle
+# of someone's browser suite — that is the common shape of the contention, and the shape #847's guard
+# does not cover, because a pre-commit gate is not a browser gate. A refusal here would be a worse tax
+# than the flake: a blocked commit costs more than a slow one, and the person committing is usually not
+# the person who can decide to wait. So this only tells you what you are about to walk into.
+if [ -z "${CI:-}" ] && [ -f "$root/scripts/lib/e2e-concurrency.sh" ]; then
+  # shellcheck source=../scripts/lib/e2e-concurrency.sh
+  . "$root/scripts/lib/e2e-concurrency.sh"
+  live_gates="$(rask_other_e2e_runs | tr '\n' ' ')"
+
+  if [ -n "${live_gates// /}" ]; then
+    echo "pre-commit: HEADS UP — a browser E2E gate is running on this machine right now." >&2
+    for pid in $live_gates; do
+      elapsed="$(ps -o etime= -p "$pid" 2>/dev/null | tr -d ' ')"
+      cmd="$(ps -o command= -p "$pid" 2>/dev/null | cut -c1-120)"
+      [ -n "$elapsed" ] && echo "            pid $pid, running for ${elapsed}: $cmd" >&2
+    done
+    echo "            This gate builds and tests the whole solution, so it will compete with that" >&2
+    echo "            suite and may make it fail for reasons that have nothing to do with either" >&2
+    echo "            change. Not refusing — a blocked commit is worse than a slow one — but if that" >&2
+    echo "            run goes red, this is the first thing to suspect." >&2
+    echo >&2
+  fi
+fi
+
 echo "pre-commit: running the local format + unit gate (staged code changes detected)."
 echo "            build -> dotnet format --verify-no-changes -> tests. bypass with"
 echo "            'git commit --no-verify' or RASK_SKIP_UNIT=1."
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -94,7 +94,7 @@ elif ! git rev-parse --verify --quiet origin/main >/dev/null; then
     "The code the CLI writes doesn't compile." \
     "A scaffolded project built, but a gate assertion failed — read the failure above."
 elif ! git diff --name-only origin/main...HEAD | grep -E "$generator_paths" >/dev/null; then
-  : # nothing in this push can change what the generators emit
+  echo "pre-push: CLI build gate SKIPPED — nothing in this push matches the generator paths."
 else
   echo "pre-push: this push touches the generators or the packages they emit — running the CLI build gate."
   echo "          bypass with 'git push --no-verify' or RASK_SKIP_CLI_BUILD_E2E=1."
@@ -111,14 +111,18 @@ fi
 # src/Rask.Core/Resources/ is in scope as a whole. The client branch used to live entirely in each
 # host's own file, so naming those two was enough; both hosts now IMPORT the indicator, the morph and
 # the diff codec from there, and an edit to any of them changes what an edit does to a running app.
-watch_paths='^(src/Rask\.Core/(HotReload|ScopedAssets|ScopedCss|Live|Routing|Resources)/|src/Rask\.Server/(LiveSessionStore|RaskEndpointExtensions)\.cs|src/Rask\.Server/Resources/rask\.ts|src/Rask\.Wasm/(HotReload/|Resources/rask\.wasm\.ts|Browser/main\.ts)|src/Rask\.(Generators|Cqrs\.Generators|Generators\.Shared)/|src/Rask\.Cli/(Commands/DevCommand|DevTarget|IBrowserLauncher)\.cs|tests/Rask\.Cli\.Tests/WatchHotReloadE2ETests\.cs|tests/Rask\.Examples\.E2E\.Tests/WasmWatchHotReloadTests\.cs)'
+# src/Rask.Core/build/ is in scope because Rask.Core.targets is where @(Watch) is BUILT: `dotnet watch`
+# collects @(Compile), @(EmbeddedResource), the project file and Razor content and never @(None), so a
+# scoped asset reaches the watcher only through the items that file adds. A change to the watch list
+# itself was the one hot-reload change this filter did not select (found while fixing #862).
+watch_paths='^(src/Rask\.Core/(HotReload|ScopedAssets|ScopedCss|Live|Routing|Resources|build)/|src/Rask\.Server/(LiveSessionStore|RaskEndpointExtensions)\.cs|src/Rask\.Server/Resources/rask\.ts|src/Rask\.Wasm/(HotReload/|Resources/rask\.wasm\.ts|Browser/main\.ts)|src/Rask\.(Generators|Cqrs\.Generators|Generators\.Shared)/|src/Rask\.Cli/(Commands/DevCommand|DevTarget|IBrowserLauncher)\.cs|tests/Rask\.Cli\.Tests/WatchHotReloadE2ETests\.cs|tests/Rask\.Examples\.E2E\.Tests/WasmWatchHotReloadTests\.cs)'
 
 if [ "${RASK_SKIP_WATCH_E2E:-}" = "1" ]; then
   echo "pre-push: RASK_SKIP_WATCH_E2E=1 set — skipping the watch hot-reload gate."
 elif ! git rev-parse --verify --quiet origin/main >/dev/null; then
   echo "pre-push: no origin/main to diff against — skipping the watch hot-reload gate."
 elif ! git diff --name-only origin/main...HEAD | grep -E "$watch_paths" >/dev/null; then
-  : # nothing in this push can change the hot-reload loop
+  echo "pre-push: watch hot-reload gate SKIPPED — nothing in this push matches the hot-reload paths."
 else
   echo "pre-push: this push touches the hot-reload loop — running the watch gate."
   echo "          bypass with 'git push --no-verify' or RASK_SKIP_WATCH_E2E=1."
@@ -146,7 +150,7 @@ if [ "${RASK_SKIP_DEPLOY_E2E:-}" = "1" ]; then
 elif ! git rev-parse --verify --quiet origin/main >/dev/null; then
   echo "pre-push: no origin/main to diff against — skipping the deploy gate."
 elif ! git diff --name-only origin/main...HEAD | grep -E "$deploy_paths" >/dev/null; then
-  : # nothing in this push touches the deploy path
+  echo "pre-push: deploy gate SKIPPED — nothing in this push matches the deploy paths."
 else
   echo "pre-push: this push changes the deploy path — running the deploy gate against a container host."
   echo "          bypass with 'git push --no-verify' or RASK_SKIP_DEPLOY_E2E=1."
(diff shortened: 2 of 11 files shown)


pal-tamas:
A live instance of the failure mode this PR is about, hit while gating an unrelated branch a few minutes ago:

```
Failed Rask.Example.Shared.Tests.Demos.LiveTickerTests.PollLoop_HistoryStaysCappedAt60 [668 ms]
  System.TimeoutException : WaitFor timed out after 00:00:20: the history never filled to capacity
```

The branch under test was the CLI dependency-floor work — `DoctorCommand`, `NodeRequirement`, two
installer scripts. There is no causal path from any of that to a polling loop in a sample app.

Re-run alone immediately afterwards: **9/9 green, and the same test completed in 3 s** against the
20 s timeout it had just exhausted. So it was starved rather than broken.

Two things worth recording:

1. This is the shape [#850](https://github.com/pal-tamas/rask/issues/850) describes — a timing-sensitive
   test that reds under load and reads as a real defect. Without the re-run it would have cost an
   investigation into `LiveTicker`.
2. It also shows why the hint had to go on the **failure** rather than the start. The competing load
   here was not a browser gate, so the existing guard had nothing to say about it; and by the time the
   suite went red the contention was still there to be named.

The run that produced it was on `origin/main`'s `run-unit-local.sh`, which is why no hint appeared —
this branch's version is what would have printed one.

Also worth noting for whoever reads the gate output: the run emitted dozens of `Test Run Successful.`
lines (one per assembly) and still exited 1. Per-assembly success lines are not the gate's verdict.

pal-tamas:
A second and third instance, from the next gate run on the same unrelated branch — and this time the competing load was directly observable.

That run failed on two different tests from the first one:

```
Failed Rask.Server.Tests.WebSockets.AsyncValidationDispatchTests.AsyncValidator_PostHandlerFrame_ShowsMessage_AndNoIndicator [4 s]
Failed Rask.Example.Shared.Tests.Guides.DemoMarkupGoldenTests.NoDemoSkeleton_ChangesOnATimer [1 m 4 s]
```

Both passed alone immediately afterwards — the WebSocket test in **2 s**, the timer golden in 47 s
against the 64 s it had just taken while starved.

What makes this one useful is that the cause was still on the process table when the gate went red:

```
28755  01:38  dotnet publish samples/Rask.Example.Playground -c Release -nodeReuse:false --nologo
29107  00:51  dotnet build Rask.slnx -c Release -m:1 --nologo -v q
```

Two other sessions, neither a browser gate — so `rask_other_e2e_runs` had nothing to say about either,
which is exactly the gap this PR is for. `rask_other_heavy_builds` matches both (`dotnet publish` and
`dotnet build` are in its verb list), so the branch's `run-unit-local.sh` would have printed their pids
and elapsed times beneath the failure and asked for a re-run alone.

Three failures across two runs, on three different timing-sensitive tests, none of them reachable from
the change under review. Without the re-runs each one is an afternoon.

Worth adding to the record for whoever tunes this later: the failures were **not** repeatable across
runs. Run one lost `LiveTicker`, run two lost a WebSocket dispatch test and a timer golden. A hint tied
to specific known-flaky test names would have missed all three; keying on "was the machine busy" is
what catches them.

pal-tamas:
One more, and this time the guard's blind spot bit the tooling written *for* this PR.

Two of my own pushes were queued behind each other, both waiting on `rask_other_e2e_runs`. The first
cleared its browser gate and moved on to the **CLI build gate** — at which point the second waiter saw
no browser gate, concluded the machine was free, and started its own browser suite:

```
33205  01:11  .../worktrees/localization-decision/scripts/run-cli-build-e2e.sh
33394  00:34  .../worktrees/scoped-asset-watch-test/scripts/run-e2e-local.sh
```

`run-cli-build-e2e.sh` packs 15 packages and runs several full builds, so this is exactly the "heavy
build competing with a live browser suite" case — and nothing refused, because a CLI build gate is not
a browser gate.

Two things worth taking from it:

1. It is more reachable than it looks. A push that clears E2E is *still* holding the machine for the
   path-filtered gates behind it, and any queue keyed on `rask_other_e2e_runs` will start the next run
   into that. The window is minutes wide, not seconds.
2. `rask_other_heavy_builds` in this PR does match it — `run-cli-build-e2e.sh` shells out to
   `dotnet build` and `dotnet pack` — so a suite that reds under this would now say so. That is the
   hint doing its job on a case I did not construct for it.

Not proposing a change here: turning the *guard* into "refuse if any heavy build is running" is the
option the issue rejects, and rightly — `dotnet` runs for dozens of legitimate reasons. But it does
argue the hint is the right layer, since it costs nothing to be right about a case the refusal cannot
safely cover.

PR #912, opened 2026-08-31
feat(tailwind)!: Tailwind is built in — no flag, no package, no off switch

pal-tamas:
## Summary

Tailwind becomes the way a Rask app is styled, rather than one of three things it could be: the
styling axis is gone from `rask new`, `Rask.Bootstrap` is deleted, and the Tailwind compiler ships
**inside** the host packages so a scaffolded project references nothing and can't turn it off.

Three things make up "built in", and the first two are new here:

- **`src/RaskTailwindBuildPack.targets`** packs the Tailwind props, targets and engine-resolution task
  into `Rask.Server` and `Rask.Wasm`, each importing them from its own
  `build/<PackageId>.{props,targets}` — the arrangement `Rask.Core`'s scoped-CSS integration already
  used. A scaffolded `.csproj` names no styling package, and an existing app gets Tailwind on its next
  host upgrade with nothing to edit. `Rask.Tailwind` is now `IsPackable=false` (both CI workflows stop
  packing it): an app referencing a host package *and* `Rask.Tailwind` would import one targets file
  twice and run the compiler twice over a single output.
- **`RaskTailwindBuild` is deleted** — from the props, from the gate, and from the three resolver
  errors that used to offer it as the way out. Those now name the way *through* (install Node, seed
  the per-user cache, unset `RaskTailwindOffline`), which between the standalone and npm engines is
  always available. What remains in the gate is a design-time check and `Exists($(RaskTailwindInput))`;
  the second is load-bearing now that the targets reach every project in a solution, and is not a
  switch, because a project that *has* a stylesheet cannot decline to compile it.
- `--bootstrap` / `--tailwind` are retired through `RetiredFlagError` (refused, not ignored), the
  `Styling` enum is gone, and the wizard drops a question.

### The showcase was shipping with no CSS

Getting the browser gate green turned up two live bugs that no assertion over text or DOM structure
could see:

- **No Tailwind was being served at all.** `App.cs` linked `/css/app.css`, but the sheet compiles into
  `Rask.Example.Shared` — a class *library*, whose `wwwroot` publishes under
  `_content/Rask.Example.Shared/`. The link 404'd on both hosts and every page rendered as unstyled
  HTML. This is exactly the failure mode `docs/tailwind.md` cites as the reason there is no off switch.
- **The brand palette resolved to nothing.** `global.css` still declared that the violet tokens "now
  live in `_content/Rask.Bootstrap/tokens.css`" — deleted with the package — so `--accent`, `--ink`,
  `--line` and `--panel-2` were undefined. The palette is now a Tailwind `@theme` in the showcase's own
  `Styles/app.css`, matching the marketing site's, which `tokens.css` itself named canonical.

Plus three behaviour regressions from the restyle: the sidebar filter stopped filtering as you type
(`BsInput` wired its handler to `OnInputAsync` internally; the raw `Input` kept `.OnChange`, which is
the DOM `change` event and fires on commit), the pinned filter scrolled away with the list (its
flex-column rule targeted `.side-nav .offcanvas-body`, a wrapper that no longer renders), and the
Todos dialog had a single exit (its comment described a sibling backdrop that was never rendered, and
the native `<dialog open>` replacing `BsModal` is non-modal — no focus trap, no Escape).

The toaster demo is restored rather than left deleted: `IToaster` and `ToastOutlet` are framework
features that survived, and only the `Bs*` chrome went — so by the rule the rest of the sweep followed
(subject deleted, chrome converted) it was a conversion. Its guide section had been left with the
sentence that introduces it and no demo after it.

E2E locators are re-pinned on things a restyle cannot quietly remove: accessible names where there
were none (the theme toggle, the todo row's two icon-only buttons and the master-detail expander were
all announcing as a bare "button"), and contract classes named for what they are — `alert`, `badge`,
`ops-nav-link`, `drawer-backdrop`, `app-dialog`, `todo-list` — rather than for the framework that used
to supply them, each commented as a contract at its call site the way `.term` already was.

## Testing

- Unit: `dotnet test Rask.slnx --filter "FullyQualifiedName!~Rask.Examples.E2E"` — **green**. (One
  `SqliteConcurrencyStressTests` 500-writer flake seen once; passes on re-run and is nowhere near this
  change.)
- E2E: `scripts/run-e2e-local.sh` — **63/63**, hosts Server, WASM, standalone WASM, Shop, Site,
  Playground and EF Core. It was red on 14 before this branch was touched.
- `dotnet format --verify-no-changes` clean; `dotnet build -warnaserror` clean (0 warnings).
- The off switch is proved dead by building a sample with `-p:RaskTailwindBuild=false` and reading the
  24 KB stylesheet back, not only by a structural assertion.
- New `TailwindBuiltInTests` pins the packed route, the absence of a scaffolded reference and the
  absence of an off switch. Each guard was proved by breaking it first, since every defect it covers
  lives on the far side of a `dotnet pack` and is invisible in-repo.
- `PackagingContractTests` no longer hard-codes `RaskCoreBuildPack.targets` as the only build pack —
  it finds the pack that ships the project in question, which is what made it fail the day a second
  pack appeared.
- Two stale `CliBuildE2E.FeedPackages` entries removed: `Rask.Bootstrap` outlived its project by one
  list (packing a deleted csproj fails the assert outright, and that gate is opt-in so nothing said
  so), and `Rask.Tailwind` joins `Rask.Core` as deliberately absent.

## Benchmarks

n/a — no render hot-path code changed. The diff is MSBuild packaging, CLI scaffolding, sample markup
and tests; `HtmlSerializer`, `Element`/`Component` render, the diff codec and the WS/WASM dispatch are
untouched. The Tailwind compiler has no runtime assembly, so nothing enters the render path or the
published bundle.

## CHANGELOG

`[Unreleased]` entries added under **Changed** (Tailwind ships inside the host packages; the off
switch is removed; `rask new` has no styling axis), **Removed** (`Rask.Bootstrap`), and **Fixed** (the
showcase serving no stylesheet, the unresolved palette, the sidebar filter, the pinned filter, the
dialog backdrop, the unclickable expander, the restored toaster demo).


pal-tamas:
Superseded by #914, which landed the union of both branches as one squash.

They shared 17 commits — two overlapping finishes of the same epic rather than a stack — so merging both would have applied near-duplicate work twice. #914's branch carried the newer restyle and the `NoBootstrapClassesTests` gate; this PR's unique commit, `feat(tailwind)!: Tailwind ships inside the host packages, with no off switch` (9f509230), was cherry-picked on top, so nothing from here is lost.

Landed as 4982a483.

PR #935, opened 2026-09-01
feat(dashboard)!: rebuild the operator console light, mobile-first, on an internal component kit

pal-tamas:
## Summary

`/_rask` worked and looked like a debug page. It now follows the design language it was specified
against: a near-white surface ladder, hairline rules, monospace for every machine value, and a
**breadcrumb bar over an underlined tab bar**.

Tailwind v4 only, zero JavaScript and a reusable component kit were **already true** of the console and
stay true, so this is a visual and information-architecture change rather than a re-platforming. `Ui/`
now holds the shell, top bar, tab bar, metric row, detail sheet, toast, search and button — all
`internal`, so the only public surface added is six `OpsIconName` members.

### The IA change

Every registered queue used to get its own top-level tab, so a deployment running jobs, outbox and mail
spent half its navigation on them and the nav grew with the batteries instead of describing the console.
They share one **Queues** section now, with the individual queue on a breadcrumb switcher — a real
`<select>`, because a menu is a popover and a popover is a key listener, and this ships no script. It is
keyboard-navigable for free and opens the platform picker on a phone. **Every existing URL still
resolves.**

### The rest removes duplication rather than adding surface

- The five counts were a tile row **and** a tab strip carrying the same numbers. One row now both reports
  and filters; each tile is a real link carrying `?show=`.
- A row's detail was an extra `<tr>` spliced underneath, which cramped a stack trace into a table column
  and pushed every row below it down on a polling page. It is a sheet with dotted-leader rows.
- The overview printed two tiles per queue; it prints one card.
- The cache's `?q=` filter has shipped since that page did, and its own empty state names it — but
  nothing ever rendered a box to type it into. Now it does.

### Mobile-first

A different claim from "responsive": secondary table columns collapse below `sm` and fold under the
primary cell rather than scrolling sideways, because a table you have to swipe has hidden the column you
came for. Controls take a 44px touch target below `sm`.

## Two defects found by looking at 390px, not by a test

1. An odd tile in a two-column row left an **empty cell that rendered as a grey block** — the hairlines
   are the container's own background showing through the `gap-px`, so an empty slot is not blank.
2. An unbreakable request id in a log scope pushed the table **wider than the phone**, hiding content
   behind the table's own `overflow-x` while the document-level overflow check stayed perfectly green.

The second is the interesting one: the element that overflows is not the element that scrolls.
`ShopExampleTests` now measures the document **and every table** against the viewport at 360px.

## Testing

- `scripts/run-unit-local.sh` — **407** unit tests, format clean.
- `tests/Rask.Dashboard.Tests` — **71** (5 new in `DashboardChromeTests`, asserting through selectors
  rather than substring matches on HTML, which pass on a class name or on a page that failed to render).
- `scripts/run-e2e-local.sh` — **64** browser tests, including the new
  `The_dashboard_fits_a_phone_without_scrolling_sideways` across all five console pages.
- **The new mobile gate was proven by failing it**: removing the log-table clamp and re-running reported
  a wide table on exactly the logs page while the page-level check stayed `ok`.
- Driven for real against `samples/Rask.Example.Shop` and screenshotted at 1280 and 390 via a new
  committed `dashboard-driver.cs` (documented in the `run-rask` skill).

Benchmarks are unaffected — this is package UI, not render hot path — and the pre-push benchmark gates
(both payload-bytes baselines plus the capacity smokes) passed unchanged.

## Notes

- `SystemPageBackupTests` asserted on `text-red-400` / `text-amber-400`. The intent — Unknown renders as
  warn, never as danger — is unchanged and now reads `text-ops-danger` / `text-ops-warn`, a semantic
  contract rather than a palette coincidence, which is why it broke.
- The compiled stylesheet grows **18.6 KB → 25.6 KB**, still embedded and inlined, still asset-free.
- `llms.txt` claimed the dashboard "adds `BsStat` to Rask.Bootstrap" in two places. `Rask.Bootstrap` is
  deleted; removed.

## Breaking change

`DashboardLayout`'s constructor takes a `Navigator`, so the queue switcher can navigate on change. A
public signature change on a type no application constructs itself.


pal-tamas:
## Review pass — 10 findings, 9 fixed here, 1 filed

Ran `/security-review` and `/code-review high` over the diff before merging.

**Security: clean.** No new XSS path — the only `Raw` in the package is the build-time stylesheet, and
every data-bearing string (payloads, errors, cache keys, log messages and scope values) reaches the DOM
through the children indexer, which encodes. `[Authorize]` is still on the layout with all six pages on
its route chain, the queue switcher's slug is allow-listed before `NavigateTo`, and every data-derived
attribute value goes through `AppendEncoded`.

**Correctness review found one thing I'd call a blocker, and it was mine:**

`586d043e` fixes it. Delete from the detail sheet raised its confirmation *underneath* the sheet — the
prompt renders in normal flow, the sheet is `fixed inset-0 z-50` over a backdrop, and `RunAsync` never
cleared `_expanded`. The operator tapped Delete and the console did nothing observable. Retry never
showed it because Retry skips confirmation entirely. `QueueDetailSheetTests` pins it, and the assertion
was proven by reverting the fix and watching it fail on the message it was written for.

**And one that undermined this PR's own gate:** the phone-overflow test waited only for the layout's
nav, which is on screen from the first paint, so on every asynchronously-loading page it measured a
spinner — and `querySelectorAll('table')` over a spinner is empty, which is indistinguishable from "no
table was too wide". It now waits for the first load and asserts it saw at least one table across the
run, so it cannot pass by measuring nothing.

**Accessibility:** `--color-ops-warn` measured **2.13:1** as text on white and `--color-ops-ok` 3.73:1 —
in a file whose own comment demands 4.5:1 and explains why `muted` sits at 0.52. Both now have `-ink`
twins for light grounds while the vivid fills stay for the status dots and the near-black toast, where
the darker ones would fail instead. Rose already passed at 5.39:1. Computed, not eyeballed.

Also fixed: `CurrentQueue()` compared the path case-sensitively while `QueuePage` resolves its slug
case-insensitively, so `/_rask/queues/Jobs` rendered correctly with its breadcrumb and switcher silently
missing; `DashboardStylesheetTests` now pins every token rather than three of ten (which is what the
stylesheet's comment already claimed), and that immediately found `--color-ops-violet` declared,
documented and referenced by nothing; the `data-inactive="true"` filler attribute is gone; and
`whitespace-nowrap` on the narrow columns beside the `w-full max-w-0` primary cell, which had started
wrapping "72 KB" onto two lines.

**Filed rather than fixed:** #936 — the Cache and Logs search filters don't reflect into the URL,
contradicting their own "shareable link" doc comments. Not a regression (Cache had no search box at all
before this PR, and Logs has behaved this way since it shipped), and fixing it properly means a public
constructor change on `CachePage`, so it doesn't belong in a restyle.

PR #941, opened 2026-09-02
docs(architecture): an animated map of every battery, from the UI to the database

pal-tamas:
## Summary

The repository had no diagram of itself. The only picture of how anything fitted together was an ASCII sketch of the render walk in `docs/architecture/`, and the batteries were a prose table — so the shape of the thing existed only in the heads of people who already knew it. There was not a single Mermaid block, SVG or PNG diagram anywhere in `docs/`.

This adds one: **every package under `src/`**, in bands from the browser down to `app.db`, with a packet tracing one journey at a time and the mock browser window gaining a row when the resulting diff lands.

Five journeys share the map — a server round trip, the WASM takeover, the two kinds of island, the SPA lane, and the durable work that outlives the request.

It is a Rask component built from the typed SVG family, baked to `assets/rask-flow.svg`, so the site, the README and the docs share one source that cannot drift. It replaces the hero typing animation; `assets/rask-chain.svg` goes with it — an asset nothing referenced.

## What it says that the prose only implied

- There are **two** render engines, not five. `RenderEngine.Server` and `RenderEngine.Wasm` are one component model over two transports.
- Islands come in **two kinds with opposite diff semantics**. A React/Preact/Lit island brings its own renderer, so its subtree is `OpaqueSubtree` and the differ refuses to descend. A Blazor island is rendered to HTML by Rask itself, so it is diffed like any other markup — `BlazorHost` documents why marking it opaque would make it paint once and silently freeze.
- The **SPA lane is not a render mode at all**. `Rask.Spa.Hosting` references no other Rask package: no `Component`, no live session, no diff.

## The two gates are the point over time

A hand-drawn diagram goes stale silently, and is only ever found stale by somebody reading it a year later. So:

- One test walks `src/` for directories containing a `.csproj` and fails when a package has no box. **It caught `Rask.Blazor` within minutes of #937 landing.** Enumerating by `.csproj` rather than by directory also means leftover `bin`/`obj`-only directories from deleted projects cannot fail it.
- The other pins the SPA chips to the template list `rask new --template` advertises, in both directions.

Plus the four inherited from the animation it replaces: the bake matches byte-for-byte, the file is a standalone document, every theme token carries a literal fallback, and every class and keyframe is prefixed.

## Two decisions worth recording

**The path is shown by adding a highlight, never by dimming the rest.** The first version faded non-participants to 0.28, which meant that at any instant most of a forty-box map was unreadable — and every still frame of it (the README's first paint, a screenshot, a reduced-motion reader) showed the washed-out version. Emphasis is cheap; legibility is not.

**Colour resolves in two layers.** Rules paint with local `--rf-*` aliases, each defined as `var(--color-token, #literal)`. Inline, the site's tokens win and the diagram follows the theme toggle; baked standalone the literals apply. Because `prefers-color-scheme` redefines the *aliases* rather than the fills, the standalone file answers the OS theme without any rule ever overriding the page's own — which matters because an ``-loaded SVG cannot see the page's theme.

## Two defects the gates caught, not review

- The new docs page was not in `GuideCatalog`, so it would have shipped **unreachable** on the docs showcase.
- This is the first doc in the repository with an image, and the showcase rewrote `.md` links but never `` sources — so a repository-relative image resolved against the SPA route and 404'd as a **silently broken image**. Images now follow the same rule links already did, pinned in both directions.

## Testing

- `dotnet format` clean; `dotnet build -warnaserror` clean.
- Full unit gate green — zero test failures across ~50 assemblies.
- Browser E2E green — 64 journeys, zero failures. `SiteExampleTests` gains assertions that the hero renders the cycling variant, that the pinned section carries no `id` (two copies of the diagram now share one page, so the prefixing is load-bearing), and that the scenario picker ships the class change.
- Benchmark gates green — both payload-bytes baselines and the live-session capacity smokes. No hotpath code changed; this is a sample component plus docs.
- The baked SVG was rendered headlessly and inspected at both themes.

## Note for review

One red gate on the way was **not** this change: two concurrent gates raced for `.git/config`, `git config core.hooksPath` exited 255, and warnings-as-errors turned that `MSB3073` warning into a build failure. The gate printed its own "re-run alone" advice; re-running on a quiet machine gave 0 build errors.

Vue and Svelte are drawn dashed because they are still on a branch. When `worktree-vue-svelte-islands` lands, they become solid — a one-line edit in the node table.


pal-tamas:
Closing at the author's request.

PR #954, opened 2026-09-02
fix(prerender): a prerendered WebAssembly page keeps the shell that boots it

pal-tamas:
## Summary

`<RaskPrerender>true</RaskPrerender>` shipped pages that could never boot, and every guard around it
reported success. This is the first of four PRs; the landing site and docs rebuild sit on top of it.

The pass writes a document per route into the published `wwwroot` — where `index.html` is already the
shell the WebAssembly SDK has just filled with the fingerprinted import map, the integrity-pinned
preload, the `<base href>` and `<script src="main.js">`. It wrote **over** it. What shipped was a
static page carrying real markup with no way to become interactive, on every prerendered route.

The framework cannot simply re-emit those tags: the boot script comes from an `IRaskRuntimeScript`
and the WASM host deliberately registers none (the runtime boots from the page shell), while the
import map's fingerprints and integrity hashes are minted by the SDK per publish. The shell is the
only place they exist, so it is now **kept and the render spliced into it**.

| From the shell | From the rendered page |
| --- | --- |
| `<base href>`, `<meta charset>` | `<title>`, and every other `<head>` contribution |
| import map, preload, every body `<script>` | the whole `<body>` |

Singletons are resolved rather than concatenated — a browser takes the *first* `<title>`, so
appending would have left every page titled whatever the shell said, which is the search result the
feature exists to fix.

### Two more defects on the same path

- **`<base href>` was rewritten only on the root page** of a sub-path publish, leaving
  `/docs/about/` resolving `main.js` against its own directory. Every published page is rewritten now.
- **Prerendering failed outright for any app referencing the `Rask` metapackage.** The companion
  compiles for `net10.0`, so a multi-targeted dependency resolves its non-browser face, and the
  metapackage's `net10.0` face carries server-only pieces a browser app never saw. `Rask.Dashboard`'s
  `NotFoundPage` met the sample's own, and RASK040 (a warning) became an error under
  warnings-as-errors and failed the publish. The companion now builds with warnings-as-errors off,
  alongside the analyzers it already disabled.

## Why nothing caught this

Three independent blind spots, all now closed:

- Every test rendered into an **empty temp directory** — the one arrangement with no shell to destroy.
- **No sample, template or CI job had ever set the property**, so it had never run end to end.
- The "wrote no pages" warning **globbed the output for an `index.html`** — which the boot shell *is*,
  so it was non-empty whether the pass wrote every page, one page, or none, and could never fire.

The pass now prints `result written=N skipped=N` and the build reads that back, warning separately
when no summary appears at all.

## Testing

- New `PrerenderShellTests` (8 cases) covering the splice: boot machinery survives, markup replaces
  the spinner, the page's title wins, exactly one of each singleton tag, script order preserved, a
  `</body>` inside a script does not truncate the shell, and no-shell falls back to the whole document.
- Two new integration cases in `WasmPrerenderTests` that put a shell on disk first. **Both confirmed
  to fail without the splice** (verified by reverting it), which the previous suite did not.
- Full `Rask.Wasm.Tests` 276/276, `Rask.Core.Tests` prerender 7/7.
- Local gate green: `dotnet format`, `-warnaserror`, unit gate 407/407, browser E2E 70/70.

### Verified end to end, not just by unit test

`dotnet publish samples/Rask.Example.Wasm -c Release -p:RaskPrerender=true` — 20 routes planned, 5
written (15 browser-only API demos throw and are skipped by design, 3 are parameterised). Each
written page carries **both** the rendered markup and a working import map, `<base href>` and
`main.js`. Repeated with `-p:RaskPathBase=/docs`: all 5 pages, including
`routing-demo/nested/profile/`, get `<base href="/docs/">`.

## Changelog

Three entries under `[Unreleased] / Fixed`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,45 @@ them until tagged releases begin.
 
 ### Fixed
 
+- **A prerendered WebAssembly page could never boot, and every guard around it said the publish had
+  worked.** `<RaskPrerender>true</RaskPrerender>` writes a document per route into the published
+  `wwwroot` — where `index.html` is already the shell the WebAssembly SDK has just filled with the
+  fingerprinted import map, the integrity-pinned preload, the `<base href>` and
+  `<script src="main.js">`. The pass wrote over it. What shipped was a static page carrying real
+  markup and no way to become interactive, on every prerendered route.
+
+  The framework could not simply re-emit those tags. The boot script comes from an
+  `IRaskRuntimeScript` registration and the WASM host deliberately registers none — the runtime boots
+  from the page shell — while the import map's fingerprints and integrity hashes are minted by the SDK
+  per publish, so managed code has nothing to reproduce them from. **The shell is now kept and the
+  render spliced into it**: the shell keeps `<base>`, `<meta charset>`, the import map, the preload
+  and every body `<script>`; the page supplies its `<title>`, the rest of its head, and the whole
+  body. The singletons are resolved rather than concatenated, because a browser takes the *first*
+  `<title>` and appending would have left every page titled whatever the shell said — the search
+  result the feature exists to fix. With no shell present the whole document is still written, and the
+  pass says so.
+
+  Three things kept this invisible. Every test rendered into an empty temporary directory, which is
+  the one arrangement with no shell to destroy. No sample, template or CI job had ever set the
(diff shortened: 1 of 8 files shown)


pal-tamas:

PR #961, opened 2026-09-02
feat(ui): the console's component kit becomes Rask.Ui, a package all three surfaces share

pal-tamas:
## Summary

The operator console was rebuilt on a reusable component kit — and the kit stayed `internal`, so the
landing site and the docs showcase could look at it and not use it. It moves to **`Rask.Ui`**: one
vocabulary for the three surfaces the framework ships, instead of three sets of utility strings that
drift.

Second of four PRs. Stacked on #954 (the WASM prerender fix); the site and docs rebuilds sit on top.

Nineteen components **move rather than being copied**, so there is one source and nothing to keep in
step:

| | |
| --- | --- |
| Chrome | `UiShell` `UiTopBar` `UiBrand` `UiNav` `UiNavTab` `UiCrumbSwitcher` `UiCrumbSeparator` `UiTopLink` `UiMain` |
| Controls | `UiButton` `UiSearch` `UiStatusDot` |
| Data | `UiMetricRow` `UiMetric` `UiDetailList` `UiDetailRow` `UiCode` |
| Overlays | `UiModal` `UiToast` |
| Support | `UiIcon` / `UiIconName`, `UiTone`, `UiStylesheet` |

It multi-targets `net10.0;net10.0-browser`, unlike `Rask.Dashboard`, which stays server-only because
its panels read a `DbContext`. This is markup and nothing else, and the surfaces that need it most run
in the browser.

## Two API changes that came with going public

- **`Tone` is a `UiTone` enum.** A string tone has no discoverable set of values and no wrong value, so
  every misspelling silently selected the neutral branch and read on screen as the component ignoring
  the caller. Fine with one caller inside one console; a trap as public API.
- **`UiBrand` takes required `Label` and `Href`.** It named the console's own overview route and
  printed "Ops" — that coupling is what kept the kit inside one application.

## The stylesheet ships with the package

Tailwind scans the project it runs in, so a compiled library's class names are invisible to a consuming
app's build and it would emit **none** of them — leaving every component unstyled with nothing
reporting it. The kit compiles its own sheet and exposes it as `UiStylesheet.Css`.

Unlike the console's, it carries **no preflight and no `html`/`body` rules**. The console may ship a
global reset because it is a mounted application owning its own document, and it learned that the hard
way when the reset landed on host applications' pages. Inline it *before* the app's own sheet — the
`--color-ui-*` palette is how a surface re-skins the kit, and an override only wins while it is read
last.

## Naming is load-bearing, in both directions

In a markup host a bare `X` is the chain's `Build<X>` entry, so a component called `Shell` collides
with `Component.Shell` and `Nav`/`Button`/`Select`/`Search` with the Rask.Html tags — and RASK040 then
gives **neither** type an entry, across the whole compilation. Hence the `Ui` prefix.

The namespace has the same hazard from the other side: `Rask.Ui` is an enclosing namespace of every
`Rask.*` compilation, and C# resolves a simple name against enclosing namespaces *before* imports. The
samples' `Ui` class of Tailwind class-name constants became unreachable the moment the namespace
existed; it is now **`Tw`**, which is what it always was (159 files, mechanical).

## Testing

- Full build clean under `-warnaserror` on both TFMs; `dotnet format` clean.
- Dashboard **88/88**, including `DashboardStylesheetTests`, which now asserts against **both** sheets
  because that is what the console inlines — asserting against this package's half alone is how
  `sm:grid-cols-5` (written by `UiMetricRow`) would read as missing on a console that renders it
  perfectly. Two new cases pin that the kit's sheet ships and that it carries no reset.
- CLI build gate **1108 passed / 13 skipped**, run locally before pushing.
- Browser E2E **70/70**.
- Packed from a clean `obj/` and the nupkg inspected: the stylesheet is embedded in **both** TFM
  assemblies and the nuspec correctly lists no dependencies.

Public surface recorded in `src/Rask.Ui/PublicAPI/<tfm>/` (404 members).

## Four gates caught real defects

Worth reading, because two of them are invisible to `dotnet build`:

1. `PackageDependencyTests` — a new packable project owes a line in the packed README and a pack step
   in **both** `nightly.yml` and `release.yml`.
2. `GuidesTests` — every doc embedded under `docs/` must appear in `GuideCatalog.All`.
3. **`NU1101`** — the CLI gate's local feed did not pack `Rask.Ui`. `FeedCoverageTests` stayed green
   throughout: it asks which packages a *template* references, and this one arrives **transitively**
   via `Rask.Dashboard`. NuGet restores a package's dependencies too, and a scaffolded project never
   names those. Added `The_feed_is_closed_over_the_packages_its_own_packages_depend_on`, which reads
   each feed package's own `.csproj` rather than keeping a second hand-written list — verified by
   removing the entry and watching it fail.
4. **`NU5019`** — `obj/$(TargetFramework)/ui.generated.css` resolves to `obj//ui.generated.css` in the
   **TFM-less outer evaluation** every multi-targeted project also gets. `dotnet build` never notices,
   because that pass compiles nothing; `dotnet pack` starts from exactly that evaluation and fails.
   Fixed by conditioning the Tailwind properties on an inner build, so the outer pass has no input,
   skips the compile, and never contributes a resource that does not exist.

## Changelog

One entry under `[Unreleased] / Added`, including the feed-closure gap.


pal-tamas:

PR #971, opened 2026-09-02
feat(site)!: the landing page is prerendered, light, links into the docs, and ships no JavaScript

pal-tamas:
## Summary

`rask.sh` served a spinner. It is a WebAssembly app, so the first thing every visitor and every crawler
received was the boot shell — the real markup did not exist until megabytes of runtime had downloaded,
on the page whose whole job is to explain the framework.

It now sets `<RaskPrerender>true</RaskPrerender>` and is served as **78 KB of real HTML** that the
bundle then takes over. Third of four PRs; stacked on #954 (prerender fix) and #961 (`Rask.Ui`).

## Prerendering is why the rest of it changed

Not a coincidence that these landed together — each one blocked it:

- **The page needed a route.** The prerender plan is built from the registered route table, and this
  app had no `[Route]` anywhere: its root component rendered the sections itself, so the pass had
  nothing to enumerate and wrote nothing. The page moved to `HomePage` behind `[Route("/")]` and `App`
  renders the `Router`.
- **The scroll reveals had to go.** `.reveal` was `opacity: 0` until an `IntersectionObserver` said
  otherwise — on a prerendered page that ships every section invisible to anyone who never runs the
  script, including the crawler the prerender is *for*.
- **So did the hero animation.** `ChainAnimation` was 499 lines of generated SVG, and
  `assets/rask-chain.svg` was baked from it and pinned byte-for-byte by a test. **Nothing embedded that
  asset any more** — the README moved to `rask-logo.svg` some time ago. Both are deleted.

The whole `App.ts` module went with them — hero canvas, growing benchmark bars, theme toggle — and with
it the hand-written `@layer components` block that existed only to serve it. The bars were drawn by
script from a `data-` attribute, so with no script they were all zero: a chart of nothing, under the
page's central claim. They are a table now.

## Light, on the kit, linking into the docs

It uses the palette `Rask.Ui` declares, so console, site and docs are one visual language. It draws
with the kit where the kit fits — `UiMetricRow`, `UiStatusDot`, `UiIcon` — and keeps its own marketing
sections, because a landing page has no tab bar to put in a `UiNav`.

**Every feature and host card is now a link into the guide about it** (27 cards), as whole-card anchors
rather than a six-character "read more" — the card already names the thing, so the card is the link.

The kit's icon set grew 13 → 30 for this. The originals were an operations vocabulary — queues,
retries, dead letters — right for a console and almost useless for describing a framework. Heroicons v2
outline, MIT, vendored as path data, same as the rest.

## Content caught up

| Was | Now |
| --- | --- |
| "30+ compile-time diagnostics" | 60+ (`docs/diagnostics.md` defines 64) |
| "(Apple M4, .NET 10)" | Apple M4 **Pro**, .NET **10.0.5** |
| Footer CTA "▶ Open the live demo" | "Docs" — the hero's was renamed; this one was missed |
| — | cards for the operator console, islands, prerendering + render modes, durable logs, WebRTC signaling, object storage |

The bundle also drops **~2.6 MB of ICU** it never needed. That opt-in arrived with a comment describing
the *docs* app — "this app hosts the guides, and the localization guide's demo formats the same values
across four languages" — copied along with the csproj, and nothing contradicted it.

## Two fixes fell out of publishing it

Neither is visible to `dotnet build`:

- **The prerender companion inherited the wrong global usings.** It is generated as
  `{App}.Prerender.csproj`, so `Directory.Build.props` saw a `Rask.Example.*` project and injected the
  in-repo consumer usings — but the by-name exclusion sparing `Rask.Example.Site` did not match
  `Rask.Example.Site.Prerender`. It was handed `global using Rask.Validation.*;` for packages the app
  deliberately does not reference and failed with `CS0234` before rendering a page. Those conditions
  now key off a base name with a trailing `.Prerender` stripped, so a companion always sees what its
  app sees. This would have hit the docs app next.
- **`Microsoft.JSInterop` still arrives transitively** through the WASM host even though the app no
  longer references it, and a trimmed publish still fails on its `IL2104`. I removed the suppression on
  the wrong reasoning and the publish caught it; it is back, with a comment that says why.

## Prerendering has a pre-boot click gap, and the gate found it

Worth reading even if nothing changes here. The E2E journey clicked the counter three times and saw
**two**. The page is served as real HTML now, so the button exists and looks clickable *before* the
runtime has taken over — the first click lands on markup with no listener attached and is lost.

The old test hid it by accident: it waited for the `<h1>` as its boot signal, and the `<h1>` only
existed *after* boot. Prerendering made that wait return instantly.

The test now makes both phases explicit — assert the prerendered markup with **no** extended timeout
(if it needs to wait, it was not prerendered), then wait for the framework's own signal before
interacting: the shell ships `<body data-rask-root>` empty and the runtime stamps the session id on it.
Verified against `rask.wasm.js` that the morph really does sync attributes rather than assuming it.

**This is a real property of every prerendered Rask WASM page, not a test artifact**, and what the
framework should do about it for real visitors is an open design question — do nothing and document it,
replay pre-boot events, or mark interactive controls not-yet-live. Not decided in this PR.

## Guards replaced, not removed

- `SiteWasmAppFixture` required a baked scoped-JS bundle for the hero canvas. There is no bundle now,
  so it asserts instead that the published `index.html` carries **both** the markup and the boot
  script — those fail in opposite directions, and each alone still renders *something* in a browser.
- New `GuideLinkTests`: every card's slug resolves to a real file under `docs/`, and every `_blank`
  link carries `rel="noopener"`. **Verified by pointing one card at a typo and watching it fail.**
- `front-doors.test.sh`'s own extractor-guard caught `CounterCodeHtml` moving file; repointed, 13/13.

## Testing

- Browser E2E **70/70**; CLI gate 36 passed / 12 skipped.
- Published for real and inspected: 78,590-byte `index.html` with the headline, one
  `<script src="main.js">`, the import map, `<base href="/">`, the page's own `<title>`, no boot
  spinner, and 23 unique `docs/guides/` links — where before it was a shell.

## Breaking

`samples/Rask.Example.Site/ChainAnimation.cs` and `assets/rask-chain.svg` are deleted.


pal-tamas:

Issue #975, opened 2026-09-02
Route links omit PathBase, so every hard navigation under a sub-path deploy 404s


pal-tamas:
## What happens

`NavLink` writes its `Href` straight through:

```csharp
// src/Rask.Html/Components/NavLink.cs
AppendUrlAttr(sb, "href", Href.Value.ToString());
```

A generated `Routes.*` URL is the route's own path — `/guides/00-overview` — so under a sub-path deploy
the rendered anchor is **root-relative and missing the prefix**. The app is served at `/docs/`, and the
link says `/guides/00-overview`.

Clicking it in the app works, because the runtime intercepts the click and routes client-side against
its own `PathBase`. Everything that is *not* an intercepted click does not:

- open in a new tab, middle-click, cmd/ctrl-click
- copy link address, then paste it anywhere
- a crawler following the link
- any hard navigation or reload of that URL

All of them go to the origin root and 404.

Scoped-asset URLs are unaffected — those go through `LiveOptions.PathBase` and come out correct. It is
specifically route URLs that skip it.

## Why it matters

`rask.sh/docs/` is a sub-path deploy, so every sidebar link in the published showcase has this shape.
It became visible when prerendering put the anchors into the served HTML, but it is not caused by
prerendering — the same hrefs are in the live DOM today, so "copy link to a guide and send it to
someone" already produces a dead URL.

## Evidence

Publishing the showcase at a prefix and reading the emitted HTML:

```bash
dotnet publish samples/Rask.Example.Wasm -c Release -p:RaskPathBase=/docs
grep -o 'href="/g[^"]*"' .../publish/wwwroot/index.html | head
```

```
href="/guides/00-overview"
href="/guides/01-scaffold"
...
```

while the same document's scoped stylesheet is correctly prefixed:

```
href="/docs/_rask/a/8d11e1d937f8.css"
```

## Suggested direction

Prefix route URLs with `LiveOptions.PathBase` where they are rendered as an `href`, the way asset URLs
already are. `NavLink` is the obvious place; worth checking whether anything else renders a `RouteUrl`
into an attribute, and whether `RouteUrl.ToString()` itself should carry the prefix (probably not — it
is also compared against `RouteState.Path`, which is prefix-less, so prefixing at the render site keeps
the comparison honest).

A test that publishes at a prefix and asserts every emitted `href` starts with it would keep this shut.

## Notes

Pre-existing; found while prerendering the docs showcase. Related: #974.

Issue #974, opened 2026-09-02
Prerendering an app with un-prerenderable routes breaks their deep links: the SPA fallback stops being a neutral shell


pal-tamas:
## What happens

`<RaskPrerender>true</RaskPrerender>` renders every *prerenderable* route and skips the rest — a route
whose segments are not all literal, or whose page throws off a browser. That is correct on its own.

The problem is what a static host then does with the routes that were skipped. It answers an unknown
path with its SPA fallback, which is the root `index.html`. Before prerendering, that file is the boot
shell: a neutral document that boots and routes to whatever URL was asked for. **After prerendering,
the root `index.html` is a fully rendered page** — the home page — and a deep link into a skipped route
is answered with the home page's markup instead of a neutral shell. It then misroutes.

## How it was found

Enabling prerendering on `samples/Rask.Example.Wasm` (the docs showcase, 20 routes) turned a large part
of the browser E2E suite red. Only 5 routes are prerenderable; the other 15 are the browser-API demos —
media devices, fullscreen, serial, USB, HID, Bluetooth — which have nothing to bind to off a browser and
are skipped by design.

A journey that deep-links to one of the fifteen, e.g. `/eyedropper`, gets the guides index and reports
a path of `/guides/eye-dropper`.

`samples/Rask.Example.Site` does not hit this because it is a **single route**: everything it serves is
prerendered, so there is no fallback left to get wrong. That is why #971 is fine and this is not.

## Why it matters

It makes the feature unusable for exactly the applications it is aimed at. A real site has some routes
that cannot be prerendered — anything parameterised, anything browser-only — so "some prerendered, some
not" is the normal case, not the edge case. Right now that combination is worse than not prerendering
at all, because the failure is a wrong page rather than a slow one.

## Suggested direction

Keep a neutral fallback. The pass already has the shell in hand (it splices every page into it), so it
could write the unmodified shell to a well-known path — `404.html`, which GitHub Pages and most static
hosts already use, and/or a copy the host can be pointed at — so a skipped route is answered by a
document that boots and routes rather than by whichever page happens to be at `/`.

Worth deciding alongside: whether the pass should say plainly in its log that N routes will fall back,
so the mixture is visible at publish rather than discovered in a browser.

## Repro

Add `<RaskPrerender>true</RaskPrerender>` to `samples/Rask.Example.Wasm/Rask.Example.Wasm.csproj`, then:

```bash
[...]

Issue #973, opened 2026-09-02
Prerendered WASM pages look interactive before they are, and clicks in that window are lost


pal-tamas:
## What happens

A prerendered WebAssembly page is served as real HTML, so its controls are present and look clickable
immediately. The handlers are not attached until the bundle downloads, starts and takes the page over.
**Anything clicked in that window is silently lost** — no visual response, no queue, no replay.

## How it was found

`SiteExampleTests` clicks the landing page's counter three times and asserts `3`. It got `2`.

The old test hid this by accident: it waited for the `<h1>` as its boot signal, and the `<h1>` did not
exist until the app had rendered it. Prerendering makes that wait return instantly, so the clicks
started racing the runtime.

The test now waits for the framework's own "I have taken over" signal before interacting — the shell
ships `<body data-rask-root>` empty and the runtime stamps the session id onto it — but that only fixes
the *test*. A real visitor gets no such wait.

## Why it matters

The window is short on a fast connection and seconds on a slow one, and it is worst exactly where
prerendering is most valuable: a first-time visitor on mobile data, on the page that is trying to
convince them. `rask.sh` ships a counter in its hero whose entire job is to be clicked.

## Options, roughly in increasing order of work

1. **Document it.** Say plainly in `docs/prerendering.md` that a prerendered page is not interactive
   until boot, and let application authors decide.
2. **Mark the state.** Have the prerendered document carry an attribute the runtime clears on mount, so
   an app can style not-yet-live controls (a cursor, a dimmed state, `inert`) with no script of its own.
   This fits the framework's zero-JS posture.
3. **Replay.** Have the boot shell record pre-boot events on the root and replay them once the runtime
   mounts. Most seamless, most machinery, and needs care about which events are safe to replay.

Option 2 looks like the best fit — it is a rendering concern, costs no JavaScript in the app, and turns
an invisible failure into something an author can address deliberately.

## Repro

```bash
dotnet publish samples/Rask.Example.Site -c Release   # <RaskPrerender>true</RaskPrerender>
```

Serve the published `wwwroot`, throttle the network, and click the hero counter as soon as it appears.
Clicks before the bundle boots do not register.

## Notes

Found while prerendering the landing site (#971). Not a regression — prerendering is what introduces
the window — but it is new behaviour that ships with the feature.

PR #977, opened 2026-09-02
feat(docs): the showcase is light on the shared kit, and finally says how to install Rask

pal-tamas:
## Summary

`rask.sh/docs/` opened straight into guide cards, so a visitor who landed there was never told how to
get a project — the marketing site had the commands and the documentation site did not. The guides
index now opens with the installer, `rask new MyApp` and `rask dev`, plus the Windows one-liner and the
`--template wasm` / `--auth` switches.

Fourth PR; stacked on #954, #961 and #971.

## The two install copies are pinned to each other

They cannot share a constant: the landing site deliberately references none of the showcase, being a
size-tuned marketing bundle that just dropped 2.6 MB of ICU. So `install-script.test.sh` pins this copy
alongside the other seven **and** asserts that both sites scaffold with the same command — each
carrying a URL is not the same as the two agreeing, and disagreeing is the failure that reaches a
visitor. 98 checks.

## Light, on the shared kit

The chrome is the kit's, so the console, the landing site and the docs share one top bar instead of
three near-identical ones. The light/dark toggle went with the light-only decision, which removed the
only reason `ShowcaseLayout` injected `IJSRuntime` — and orphaned its scoped `ShowcaseLayout.ts`, whose
sole export was `toggleTheme`. Both deleted.

The pre-paint theme script now stamps `light` rather than reading a saved choice and defaulting to
dark. The showcase's own stylesheet is still dark-first at `:root` and its ~170 demo pages are not
ported yet, so that attribute is what keeps them agreeing with the chrome in the meantime — one change
instead of touching 170 files.

**The sidebar stayed local** rather than being forced into `UiNav`: the kit's nav is a five-tab console
bar and this app has eighty guides in a filterable, grouped rail. Sharing chrome that does not fit
would have been worse than sharing none.

Nine generic primitives moved out of `Rask.Dashboard` into the kit — `UiCard`, `UiStat`, `UiHeader`,
`UiTabs`, `UiTab`, `UiNotice`, `UiBadge`, `UiTable`, `UiGrid` — with the class-name constants they
share, now `UiStyles`. **The console's 88 tests pass unchanged**, which is what says the move was
behaviour-preserving. Three icons added for the shell (`Menu`, `Play`, `Book`), Heroicons v2 outline.

## Prerendering the showcase is NOT here

I tried it, and backing it out is the right call. Only 5 of its 20 routes can be prerendered — the
browser-API demos have nothing to bind to off a browser and are skipped by design — which leaves the
bundle serving a **mixture**. A static host answers an un-prerendered deep link with its SPA fallback,
the root `index.html`, and once the root is prerendered that fallback stops being a neutral boot shell
and becomes a fully rendered guides index. Deep-linking into one of the fifteen then misroutes.

The landing site does not hit this because it is a **single route**: everything it serves is
prerendered, so there is no fallback to get wrong. I generalised from that without checking.

Filed as #974, with the reasoning also left in the csproj where the next reader will find it.

## One real prerender fix did fall out of the attempt

**A prerendered page under a sub-path asked the origin root for its own assets.** A browser boot reads
the URL prefix off `<base href>`; the pass has no document, so every `PathBase`-prefixed URL was baked
against an empty prefix — `/_rask/a/x.css` rather than `/docs/_rask/a/x.css`. A `<base href>` cannot
rescue those: it applies to *relative* URLs only, and a leading slash means the origin root.

The build now passes `RASK_PRERENDER_PATH_BASE`, and an explicit `PathBase` set in `Program.cs` still
wins over it, matching the browser boot's precedence. Two unit tests; confirmed by republishing at
`/docs`, where unprefixed `"/_rask/` occurrences went from 1 to 0.

This is a prerequisite for #974 whenever the docs prerender lands.

## Not here either: icon convergence

`Icon.cs` declares 132 members and **148 distinct references** exist across `samples/` and `tests/`.
Mapping them in one diff would be 148 judgement calls made blind, several with no clean Heroicons
equivalent (`Braces`, `Bricks`, `AppIndicator`, `Broadcast`). Each per-area PR converts its own pages'
icons where a reviewer can see the page the icon sits on, and `Icon.cs` is deleted by the last one — at
which point its removal is provable rather than asserted.

## Testing

- Browser E2E **70/70**; install gate **98 checks**; CLI gate green.

The showcase journey used to drive the theme toggle end to end. It now asserts the toggle is **absent**
*and* that the navbar still carries the kit's ink token — so both a toggle reappearing and the chrome
drifting back off the shared palette are caught, rather than the coverage simply being deleted.

## Follow-ups filed

- #973 — prerendered pages look interactive before they are; pre-boot clicks are lost.
- #974 — the SPA-fallback problem above.
- #975 — route links omit `PathBase`, so hard navigation under a sub-path 404s (pre-existing).


pal-tamas:

PR #980, opened 2026-09-03
refactor(docs): the showcase's whole palette moves onto the shared kit, in one file

pal-tamas:
## Summary

`Tw` is the showcase's styling vocabulary: class-name constants rather than a wrapper component per
control, deliberately, because the showcase's subject is the **framework** and a component per button
would put a layer of the showcase's own invention between the reader and the thing being shown.

That design is what makes this cheap. Re-pointing its 56 constants from raw Tailwind hues
(`violet-600`, `slate-200`, `emerald-600`) at the kit's `--color-ui-*` tokens moves **153 call sites**
onto the shared palette with **no demo page edited**, and takes every `dark:` twin with it — the
showcase is light now, so a second set of colours had nothing left to select.

Fifth PR; stacked on #954, #961, #971 and #977. Carries a merge of `origin/main`.

## What kept its own hue, and why

`Info`, and the `Light`/`Dark` pair. This showcase demonstrates a full range of control weights side by
side, and the kit's palette is deliberately smaller than that: it names the states an operator surface
needs (ok, warn, danger) rather than a decorative spectrum. Collapsing them onto `brand` would have
printed the same button twice in a demo whose whole point is that they differ.

The older `var(--accent)`/`var(--ink)`/`var(--panel)` layer stays, re-pointed at the kit's tokens
rather than repeating hexes. ~100 demo files still name those, and this is what let the palette change
without editing them; each per-area PR retires the ones its pages use.

## The palette must be inline, and finding that out cost two false starts

Moving it to a sibling `ui.tokens.css` and pulling it back with `@import` **compiled without error and
emitted a sheet of the right size — with zero `ui-*` utilities in it.** A `@theme` reached through an
import after the Tailwind entry points is not interpreted as theme: a literal `@theme { … }` survives
into the output, which a browser ignores, and the namespace is never registered. Every component
silently loses its colour while the build stays green.

It emptied the **kit's own** sheet as well as the showcase's. Caught by grepping the compiled CSS for
the utilities rather than trusting the build.

So the showcase declares its own copy of the tokens — it must, because Tailwind emits a utility only
where it can see the token, and `Tw` writes classes (`bg-ui-brand/10`, `ring-ui-line/40`) the kit's
compiled sheet does not contain. New `UiPaletteTests` asserts the copy is identical to `Rask.Ui`'s,
**verified by changing one value and watching it fail**.

Worth knowing: until this PR that arrangement worked only by **accident of overlap** — the landing
site's `ui-*` utilities happen to be ones the kit already emits, and the first app to reach for one it
does not would have rendered unstyled with nothing reporting it.

## Test changes, none of them a literal swap

- `FloatingLabels`/`HttpPage` pinned `bg-emerald-50` and `bg-red-50`. They assert via the `Tw`
  constants now, so they pin *"this is the success/danger alert"* rather than *"this is currently
  emerald"*.
- `ShowcaseLayoutTests` pinned `bg-slate-900`, under a comment calling it "what makes it the dark bar"
  — a decision since reversed. It asserts `bg-ui-bg`: that the chrome takes its surface from the shared
  palette rather than inventing one.
- The showcase journey pinned `--accent` to one of two hard-coded violets. It now reads **both** custom
  properties in the browser and asserts `--accent` resolves to `--color-ui-brand` — the actual
  contract, so a deliberate re-skin will not turn it red. It also asserts non-empty, because two
  *undefined* custom properties both read as `""` and would otherwise compare equal, which is precisely
  the "palette silently gone" failure the step exists to catch.
- `DemoMarkup.golden.txt` regenerated through its documented `RASK_UPDATE_GOLDEN` path, and the diff
  **read** rather than accepted: 627 lines before and after, so structure is untouched and only colour
  classes changed. A line-count change would have meant markup moved, not palette.

The four `Rask.Dashboard` stylesheet failures in the middle of this were that guard doing its job — it
asserts on the **shipped artifact**, so it caught the emptied sheet immediately. 88/88 again.

## Merging main

Only `CHANGELOG.md` conflicted; both sides were legitimate `[Unreleased]` entries, so they are merged
section by section rather than one winning. Three hazards that were **not** conflicts:

- `IslandsDemo.cs` arrived from #958 using `Ui.Card`. That holder is `Tw` here — `Rask.Ui` is a
  namespace now, and C# resolves a simple name against enclosing namespaces before imports, so a type
  called `Ui` under `Rask.` is unreachable. New code on main reintroduced the collision.
- `Rask.External.Tasks.dll` in `build/` was **stale** (28 KB) against main's targets passing a new
  `PreactConfigPath` — `MSB4064` before a line compiled, and the task project could not rebuild itself
  because building it loads that same DLL. Removed and regenerated (48 KB).
- `node_modules` predated Solid islands, so `tsgo` could not resolve `solid-js`.

**Content re-verified against the merged tree** rather than notes taken before it: islands are **seven**
runtimes now (React, Preact, Solid, Vue, Svelte, Angular, Lit), not the four the landing page claimed,
and `Rask.Meta.Hosting` is a subsystem it did not mention at all. Both corrected in the feature grid;
the new card links to the real `meta` guide slug, so `GuideLinkTests` covers it.

## Testing

- Unit suite **6,779 passed / 0 failed across 51 projects** (up from 6,671/50 — main's new tests pass).
- Browser E2E **77/77**; CLI gate green.

One E2E run failed before this on `/guides/http-and-files`, an assertion **new from main's #976**. It
was my own stale state: the publish directory still held four prerendered sub-pages from earlier
`RaskPrerender` experiments, some carrying `<base href="/docs/">`, because a publish reuses that
directory rather than cleaning it. Cleared, and the suite is 77/77. Not a bug in main.


pal-tamas:

PR #982, opened 2026-09-03
refactor(docs): the showcase's demo markup drops its own colours for the shared palette

pal-tamas:
## Summary

**281 replacements across 137 files**, and 245 of them are one idiom: `text-slate-500
dark:text-slate-400` — muted secondary text spelled as a light/dark pair, which is exactly
`text-ui-muted`. Every `dark:` variant is now gone from the showcase's demo markup: the showcase is
light, so the second half of each pair selected nothing.

Sixth PR; stacked on #954, #961, #971, #977 and #980. #980 moved the *vocabulary* (`Tw`'s 56
constants) onto the kit; this moves the colour written directly into demo markup, which `Tw` never
covered.

## Scope mattered more than the substitution

That idiom appears in **145 files across ten sample apps**. Only the **125 in the showcase trio** are
converted — `Rask.Example.Shared`, plus the two hosts whose `Features` render through Shared's
stylesheet.

The standalone samples (Auth ×4, EfCore, Shop, Sqlite) compile their **own** `app.css`, which declares
no `--color-ui-*` tokens. A `ui-` utility there would be emitted by nothing and render unstyled — the
same rule that has already bitten twice on this branch: **Tailwind emits a utility only where it can
see the token.**

## Left alone deliberately, because the colour *is* the demo

- The Context **theme badge** — it demonstrates a dark/light badge.
- The dark `<pre>` **code blocks**.
- **Gantt's** progress hues.
- `Tw`'s **Info** variant, which the kit has no token for and which that file documents.

## Two real fixes in the tail, not mechanical swaps

- The **field-error red** (`text-red-600 dark:text-red-400`) becomes `text-ui-danger`, four sites.
- **Orphaned `dark:` halves** the pair rules stranded — `dark:divide-slate-700 dark:ring-slate-700` and
  a `dark:bg-slate-800` dialog surface. Their light halves had been converted separately, leaving dead
  classes pointing at a theme that no longer exists.

## The golden test is the evidence

`DemoMarkup.golden.txt` renders every demo and diffs its class skeleton, so this is checkable rather
than eyeballed. Regenerated through its documented `RASK_UPDATE_GOLDEN` path:

- **236 lines changed, with equal add and remove** — no markup moved.
- **Zero** added lines still contain `slate-` or `dark:`.

A line-count change would have meant the pass edited structure rather than colour.

Two tests pinned the old classes and now pin the new: `AppTests` on the body class, `PageHeaderTests`
on a subtitle.

## Testing

- Unit suite **6,779 passed / 0 failed across 51 projects**.
- Browser E2E **77/77**; CLI gate green.

One note for the record: the first commit attempt failed on
`Rask.Meta.Hosting.Tests.SupervisorSeamTests`, a test in main's new subsystem that a palette change
cannot reach. Re-run alone it is 46/46, and the retry passed. Load-sensitive rather than caused here —
but it is the second flake seen from that area, and worth an issue if it recurs.

## What is left after this

Only the icon convergence: `Icon.cs`'s 132 members and 148 references. That is the judgement-heavy
piece — several Bootstrap-shaped names (`Braces`, `Bricks`, `AppIndicator`, `Broadcast`) have no clean
Heroicons equivalent — so it stays per-area, with `Icon.cs` deleted by the last PR once nothing
references it.


pal-tamas:

PR #1000, opened 2026-09-04
feat(gates): the browser E2E gate queues for the machine instead of refusing it

pal-tamas:
## Summary

The browser E2E gate now queues for the machine's exclusive slot instead of
refusing to run when another job holds it, so a parallel session no longer turns
a green branch into a skipped gate.

## Testing

⚠️ **The ship gate has NOT been re-run on this branch** while closing out
background sessions. Run `scripts/run-unit-local.sh` and
`scripts/run-e2e-local.sh` before merging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,41 @@ them until tagged releases begin.
   its own test. The identity now travels in `GIT_AUTHOR_*`/`GIT_COMMITTER_*`, which cannot leak into a
   config file, and four assertions pin `HEAD`, the working tree, the committing identity and the
   absence of any `git config user.*` write.
+### Changed
+
+- **The browser E2E gate queues for the machine instead of refusing it.** The lane is a machine-wide
+  singleton — two suites contend and the contention surfaces as a plausible-looking red minutes later
+  — so finding another gate running used to fail the push with "wait for the run above to finish, then
+  push again". That was right about the contention and wrong about what to do with it: every caller
+  hand-rolled a waiter, and anyone who did not got a failed push for a machine state that had nothing
+  to do with their branch. The gate now names the run holding the lane and **waits for it**, starting
+  automatically when it frees.
+
+  The serialization guarantee is unchanged: still exactly one suite at a time. What went away is the
+  manual retry.
+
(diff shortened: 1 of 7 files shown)


pal-tamas:
Closing as **superseded** — every behaviour this branch adds is already on `main`, and nothing on it is unique.

This PR branched from `5ce3a94f` (#930). Since then `main` landed #934 and then #969, which replaced the single browser lane with the 10-slot machine budget in `scripts/lib/machine-lane.sh` + `scripts/lib/lane-claim.sh`. This branch adds queueing to the design #969 removed.

Point by point:

- **`scripts/lib/e2e-concurrency.sh`** — the branch's sole addition to this file is `rask_e2e_lane_holders`, which is the exact function `main` deliberately deleted and re-homed as `rask_lane_senior_gates` in `machine-lane.sh`, carrying the age-ordering, FIFO and tie-break-on-pid argument over verbatim. `main` even leaves a tombstone comment where it used to live.
- **`scripts/run-e2e-local.sh`** — `main` already queues rather than refuses: `rask_e2e_await_slots` polls `rask_lane_fits`, defaults `RASK_E2E_QUEUE=1`, `RASK_E2E_QUEUE_TIMEOUT=5400`, `RASK_E2E_QUEUE_POLL=20`, and refuses only under `RASK_E2E_QUEUE=0`. It is strictly better than this branch: it splits the gate by phase, so the build and publishes take a partial claim and only the browser suite is exclusive, and it publishes the claim *before* waiting.
- **`scripts/lib/build-failure.sh`** — `main`'s busy matcher is byte-identical to the one proposed here, including the "deliberately NOT matched on the holds-this-machine banner" comment. Both trees carry the same 11 `busy` references.
- **`docs/development-workflow.md`** — `main` already documents the queue env vars and the age-ordering rule.
- **Tests** — `main` has `scripts/tests/e2e-await-slots.test.sh` and `scripts/tests/machine-lane.test.sh` on top of the `e2e-concurrency.test.sh` this branch edits.

All four merge conflicts against `main` are this branch trying to re-add the single-lane design that #969 replaced, so there is nothing to salvage.

Issue #992, opened 2026-09-03
WASM guide journey flakes under machine contention and blocks pushes


pal-tamas:
`WasmExampleTests.Journey_WalksEveryPageAndUnusualActivity` fails under machine contention, at one specific assertion, and passes when run alone. It has blocked three pushes from one branch.

**The assertion** (`SharedSmokeTests.Journey.cs`, `WalkHttpAndFilesGuideAsync`):

```csharp
await Expect(Page.Locator(".guide-demo .sample-result-body article").First).ToBeVisibleAsync(
    new LocatorAssertionsToBeVisibleOptions { Timeout = 30_000 });
```

That waits on the card `HttpFetchDemo` loads with an injected `HttpClient` in `OnMountAsync` — a real fetch of `data/posts-1.json`, in WASM.

**What the failure looks like.** The aria snapshot shows the demo sitting at **"Loading…"**, not at the error banner. That is the interesting part: the component gives each attempt a 5s deadline (`AttemptTimeout`) and retries 3× before setting `_error`, so any *failing* fetch should have produced the banner well inside 30s. Sitting at "Loading…" means neither the fetch nor its own `CancelAfter` deadline resolved — consistent with the single-threaded WASM runtime being starved rather than with a transport failure.

**Evidence it is environmental.** The gate's own contention detector named the competing jobs each time, e.g.:

```
run-e2e-local: a heavy build was running on this machine during this suite.
   pid 58807 ... dotnet build Rask.slnx -c Release -m:1 -p:WasmBuildNative=false
   pid 61488 ... dotnet publish .../builtin-validation/samples/Rask.Example.Wasm
   pid 61897 ... dotnet publish .../auth-every-host/samples/Rask.Example.Wasm
```

Run alone the same journey passes in 37–40s (both the hosted and standalone variants). Under contention it fails at 50–54s. 77 of 78 tests passed in the contended run.

**Why file it rather than keep re-running.** The 10-slot budget bounds the *unit* gate and serialises the *browser* phase against other browser suites, but it does not stop another worktree's Release build or WASM publish from running beside the browser phase — which is exactly what happened all three times.

**Options worth weighing:**
- Have the browser phase wait on heavy builds too, not only on other browser suites (the detector already identifies them — it warns instead of waiting).
- Give this one assertion a budget proportional to observed load, or retry the journey once when the detector fired.
- Leave it, and treat the detector's warning as the instruction to re-run — which is what it says today, but it costs a full gate run each time.

[...]

Issue #989, opened 2026-09-03
The WASM guides journey still flakes after #976, and it fails hardest under gate load


pal-tamas:
## What happens

The WASM guides journey still fails intermittently after #976, on the same assertion that fix targeted:

```
Microsoft.Playwright.PlaywrightException : Locator expected to be visible
Error: element(s) not found
Call log:
  - Expect "ToBeVisibleAsync" with timeout 30000ms
  - waiting for Locator(".guide-demo .sample-result-body article").First
```

The aria snapshot shows the page loaded and the shell rendered — nav, sidebar, `code: /guides/http-and-files`
— with the demo's result body never filling.

Both journeys hit it, because the assertion is in the shared journey code #976 touched:

- `Rask.Examples.E2E.Tests.WasmExampleTests.Journey_WalksEveryPageAndUnusualActivity`
- `Rask.Examples.E2E.Tests.StandaloneWasmExampleTests.Journey_WalksEveryPageAndUnusualActivity`

## Observed rate

#976 measured "about one in seven" after its fix and recorded four consecutive green runs. Over one
working session on this machine I saw it block **three separate `git push` gate runs**, and:

- `StandaloneWasmExampleTests` failed in a push gate, failed once more in isolation, then passed
  **4/4** when run repeatedly on an otherwise idle machine.
- `WasmExampleTests` failed in a push gate, passed in isolation, passed in a later full gate, then
  failed again in a push gate.

So it is real, it is still there, and it is worse when the machine is loaded — which is precisely when
a push gate runs, because the gate is itself the load. That makes it disproportionately expensive: it
almost never fails when you go looking for it, and reliably fails when you are trying to ship.

## Why this is worth another look rather than a re-run

The failure mode is a **30-second timeout on a locator**, so it reads as "the page is broken" and costs
an investigation every time. Three times in one session, each costing a full gate re-run (the browser
suite is the long pole), is more than the flake budget for a demo assertion.

Two directions, both cheap relative to the cost of the current state:

1. **Wait for the demo's own signal instead of its rendered output.** The assertion waits for
   `.guide-demo .sample-result-body article` to become visible — a rendering side effect. If the demo
   exposed a completion marker (a `data-` attribute set when the fetch resolves), the wait would be on
   the thing that actually finished.
2. **Give this one assertion a longer deadline than 30s**, or skip the HTTP demo step on the WASM
   journeys specifically. #976 already established that the WASM hosts are a different timing regime —
[...]

Issue #972, opened 2026-09-02
test(e2e): the WASM guides journey fails ~1 run in 2, and it is not load


pal-tamas:
`WasmExampleTests.Journey_WalksEveryPageAndUnusualActivity` fails roughly **one run in two**, on an
idle machine, waiting for the HTTP-fetch demo on `/guides/http-and-files`:

```
Microsoft.Playwright.PlaywrightException : Locator expected to be visible
Error: element(s) not found
  - Expect "ToBeVisibleAsync" with timeout 60000ms
  - waiting for Locator(".guide-demo .sample-result-body article").First
```

## Evidence that this is not load

It first appeared during a `pre-push` run, which is the usual shape of a load-induced timeout. It is
not that:

| Run | Result |
|---|---|
| pre-push (other work on the box) | FAIL |
| isolated, attempt 1 | pass (37s) |
| isolated, attempt 2 | **FAIL** (1m26s) |

Nothing else was running for the isolated attempts. A 60-second budget is also far too generous for a
timing shortfall — the passing runs render this demo in a couple of seconds.

## Ruled out

- **Not a renamed class.** `.guide-demo` (`Shared/Markdown.cs`) and `.sample-result-body`
  (`Shared/CodeSample.cs`) both still exist. Worth stating because this repo has been bitten by a
  restyle that turned class renames into E2E *timeouts* rather than into "class renamed".
- **Not a missing asset.** `samples/Rask.Example.Wasm/wwwroot/data/posts-1.json` is present, and is
  copied into the publish output.
- **Not caused by the branch that hit it.** Found while pushing an unrelated change confined to
  `Rask.Meta.Hosting`; reproduced on `main`'s own sample.

## Where to look

`HttpFetchDemo.OnMountAsync` already carries retry handling for a documented WASM race — a
freshly-booted page's first fetch racing the discarded page's network teardown, which surfaces as
`HttpRequestException` with a null `StatusCode`. It retries 3 times, 150ms apart.

Two candidates worth separating:

1. The retry budget (3 × 150ms) is too small for the race it was written for, so the demo lands in its
   error branch — in which case the page shows the error banner and the assertion is right to fail.
2. The failure arrives as something the `when (ex.StatusCode is null)` filter does not match, so it
   falls to the final `catch` and renders the banner on the first attempt.

Either way the aria snapshot in the failure would say which, and it is worth capturing: the run that
failed showed neither the article, nor a spinner, nor an error banner in the region I inspected, which
[...]

PR #1003, opened 2026-09-04
test(e2e): give the SQLite database assertions a budget that survives gate load

### LINKED ISSUE #962: test(e2e): the SQLite bulk-import assertion has ~2s of headroom and fails under gate load
`SqliteExampleTests.Bulk_import_lands_every_row_on_both_paths` asserts that `#import-result` contains
`Change tracker:` with Playwright's default 5s expect timeout. On an otherwise idle machine the journey
completes in **1–3 seconds**, so the assertion has roughly two seconds of headroom.

That is not enough on this machine. It failed during a `pre-push` run while two other worktrees were
running their own gates:

```
Locator expected to contain text 'Change tracker:'
But was: 'Rows imported so far: 0.'
  14 × locator resolved to <div id="import-result">Rows imported so far: 0.</div>
```

Re-run three times in isolation immediately afterwards: **3/3 passed**, 3s / 1s / 2s.

## Why this is worth fixing rather than shrugging at

The E2E lane is machine-wide and queues, but the *unit* suites and the CLI build gate are not part of
that lane — so a browser journey can and does run while several other builds are saturating the box.
A five-second budget for work that takes two is a timeout that reports load as a product failure.

The cost of leaving it is not the red run, it is what a red run teaches. This repository's own
casebook records seven "flakes" that were all real bugs; a journey that cries wolf under load is
exactly what erodes the instinct to go and check.

## Suggested fix

Raise the expect timeout for this assertion specifically — the import is a bulk operation and its
duration is load-dependent by nature — rather than raising the global default, which would slow every
genuine failure down. Worth a sweep for sibling assertions on the same page with the same shape.

Found while pushing an unrelated branch (the SPA one-project layout change); nothing in that diff
touches the SQLite sample.

pal-tamas:
## Summary

Closes #962.

The SQLite journey's six database assertions ran on Playwright's 5s default, which
left roughly 2s of headroom — enough to pass on an idle machine and fail whenever
the gate is under load. They now get their own 30s budget.

## Testing

**The full pre-push gate did run on `81a088fc`.** All four `SqliteExampleTests`
passed, including `Bulk_import_lands_every_row_on_both_paths` — so the change
itself is verified.

The gate still went **red overall**, on
`StandaloneWasmExampleTests.Journey_WalksEveryPageAndUnusualActivity`, waiting on
`.guide-demo .sample-result-body article`. That is #972 and is unrelated to this
diff. Consequence: a plain re-push will keep being blocked by #972 at roughly
3-in-28, so merging needs a human who knows that the red is pre-existing.

Full gate results are in the comment thread below.

## Related

#972 / #989 / #992 track the WASM guides journey flake, which is separate and
still open.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,24 @@ them until tagged releases begin.
 
 ### Fixed
 
+- **The SQLite journey's database assertions reported machine load as a product failure.** The bulk
+  imports and the concurrent-writer bursts waited on Playwright's default 5s budget for work that
+  takes 1–3s idle, so roughly two seconds of headroom stood between a green suite and a red one
+  ([#962](https://github.com/pal-tamas/rask/issues/962)).
+
+  That is not enough on this machine. The E2E lane serialises browser journeys, but the unit suites
+  and the CLI build gate are not in it, so a journey routinely runs while several other builds
+  saturate the box — and the failure it produced (`Rows imported so far: 0.` after 14 polls) reads
+  exactly like a product bug.
+
+  The six assertions that wait on real database work now get their own 30s budget; the ones that wait
+  on a render keep the default, because raising it globally would slow every genuine failure to the
+  pace of the slowest load-dependent one.
+
+  The cost of leaving this was never the red run — it is that a suite crying wolf under load is how a
+  real regression gets waved through. This repository's own casebook records seven "flakes" that were
+  all real bugs.
+
 - **`rask new` ran a production front-end build during scaffolding.** With the batteries on, `rask new`
   creates the first migration, and `dotnet-ef` builds the project to load the `DbContext`. That build
   took `RaskSpaBuild` / `RaskMetaBuild` at their default of `true`, so scaffolding a front-end template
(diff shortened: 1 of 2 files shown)


pal-tamas:
## The gate has now run on this exact SHA (81a088fc)

Ran as a `pre-push` from a second worktree on the same commit, so this fills the gap noted above.

**The change itself is verified.** All four SQLite journeys passed, including the one this PR is about:

```
Passed SqliteExampleTests.Concurrent_writers_all_commit            [420 ms]
Passed SqliteExampleTests.Concurrent_immediate_writers_all_commit  [104 ms]
Passed SqliteExampleTests.Bulk_import_lands_every_row_on_both_paths  [1 s]
Passed SqliteExampleTests.Page_reports_wal_and_foreign_keys_pragmas [65 ms]
```

**The gate still went red, on something unrelated:** `StandaloneWasmExampleTests.Journey_WalksEveryPageAndUnusualActivity`, waiting on `.guide-demo .sample-result-body article` — #972, which is open and known to fail at roughly 3-in-28 after #993. Nothing in this diff goes near it.

So `pre-push` blocked this push, and the reason is the failure mode this PR exists to stop: a suite that reports contention as a product failure. Fixing that class in the SQLite journey does not stop #972 from doing it in the WASM one.

Worth knowing before merge: the assertion budget was verified to be actually in effect rather than silently ignored, by setting it to 1ms — exactly the three modified tests failed and the untouched pragma test passed. So a green run here is a real green, not an option the API accepted and dropped.

